PR #5, opened 2021-06-22
Add rate limit handling to gql call

lishanl:
- add throttled error handling and retry for`execute_gql`
- update the GraphQL API doc res, the gql call returns `jsondata['data']` not `jsondata` (this is different from the old spylib, I take it as an intentional change but the api doc has not reflected that, hence the fix in the doc here. )


hillairet (approved):
That's a good first version. We'll need to make this more like the rest one though where it keeps track of the credits and retries things only when there is enough credit available.

PR #8, opened 2021-06-23
:sparkles:  Adding multipass function

ponty33:
# Changes
- Added `Multipass` class to generate multipass token or URL
- Updated doc
Screenshot_20220427_162750


README.md
```diff
@@ -195,4 +195,21 @@ result of the installation and the login processes respectivaly. They are meant
...
+### Shopify Multipass
+[Shopify Multipass](https://shopify.dev/docs/admin-api/rest/reference/plus/multipass) <br>
+This helper class generates token or URL that's needed for Shopify Multipass login (Only available in Shopify Plus)
+```python
+from spylib import Multipass
+
+# Customer email is required to generate token or URL
+customer_data = {"email": "customer@email.com"}
+# Initiate the Multipass class
+multipass = Multipass("ShopifyMultipassToken")
```
hillairet: It's called a secret in Shopify's documentation so let's be consistent.
```suggestion
multipass = Multipass("MULTIPASSSECRET")
```

hillairet (commented):
@ponty33 Can this be completed? What's left?

README.md
ponty33: I think it should be ready to go.

hillairet:
Copy your README.md to docs/index.md to fix the test.
These two files should be identical.

hillairet:
@lishanl This could be useful. Do you mind looking at it to to see if it fits the rest of the code.
@ponty33 @lishanl Obviously up to you guys to decide who has time/interest in resolving conflicts and update things.

I think multipass needs its own section in the docs so it appears clearly in the menu on the left.

ponty33:
I think I can work on it, good start for me to be involved in Spylib
But wow, it has been 10 months

ponty33:
@lishanl Is there a way to preview the doc?

lishanl:
> @lishanl Is there a way to preview the doc?

start your mkdocs locally to check https://satelcreative.github.io/spylib/development-contributing/#build-and-run-documentation-lazydocsmkdocs @ponty33

ponty33:
Updated doc:
Screenshot_20220427_162750

spylib/multipass.py
```diff
@@ -0,0 +1,38 @@
...
+        self.signature_key = key[16:32]
+
+    def generate_token(self, contact_raw_data: Dict[str, Any]) -> bytes:
+        contact_raw_data['created_at'] = datetime.datetime.utcnow().isoformat()
+        cypher_text = self.encrypt(json.dumps(contact_raw_data))
+        return urlsafe_b64encode(cypher_text + self.sign(cypher_text))
+
+    def generate_url(self, contact_raw_data: Dict[str, Any], url) -> str:
+        token = self.generate_token(contact_raw_data).decode('utf-8')
+        return '{0}/account/login/multipass/{1}'.format(url, token)
```
lishanl: interesting, may update it to use f-string. 😎

docs/multipass.md
```diff
@@ -0,0 +1,21 @@
...
+
+# Initiate the Multipass class
+multipass = Multipass("MULTIPASS_SECRET")
+
+# Generate URL
+url = multipass.generate_url(customer_data, "https://{ShopifyDomain}.myshopify.com")
+# https://{ShopifyDomain}.myshopify.com/account/login/multipass/{MultipassToken}
+
+# Generate token
+token = multipass.generate_token(customer_data)
```
lishanl: just curious, in what occasion would the library user would want to return the token? I am thinking if the `generate_token` is kinda like a private / internal method that the library user doesn't need to know/use.

docs/multipass.md
```diff
@@ -0,0 +1,21 @@
...
+
+# Customer email is required to generate token or URL
+customer_data = {"email": "customer@email.com"}
+
+# Initiate the Multipass class
+multipass = Multipass("MULTIPASS_SECRET")
+
+# Generate URL
+url = multipass.generate_url(customer_data, "https://{ShopifyDomain}.myshopify.com")
+# https://{ShopifyDomain}.myshopify.com/account/login/multipass/{MultipassToken}
```
lishanl: minor thing on the casing and the quotation marks here, `'https://{store_name}.myshopify.com'` ..  just to be consistent with the rest of the doc
e.g. https://satelcreative.github.io/spylib/admin-api/#querying-shopify

docs/multipass.md
```diff
@@ -0,0 +1,21 @@
+# Multipass
+[Shopify Multipass - Shopify Documentation](https://shopify.dev/docs/admin-api/rest/reference/plus/multipass) <br>
+This helper class generates token or URL that's needed for Shopify Multipass login (This feature is only available for Plus Store)
```
hillairet: The Plus store only thing could change so let's not make our docs future proof if/when possible. Readers can find that out themselves in the docs you linked.
```suggestion
This helper class generates token or URL that's needed for Shopify Multipass login.
```

lishanl (changes requested):
Thank you Frank for adding and completing this feature 😎

docs/multipass.md
```diff
@@ -0,0 +1,21 @@
...
+
+
+# Customer email is required to generate token or URL
+customer_data = {"email": "customer@email.com"}
+
+# Initiate the Multipass class
+multipass = Multipass("MULTIPASS_SECRET")
+
+# Generate URL
+url = multipass.generate_url(customer_data, "https://{ShopifyDomain}.myshopify.com")
```
hillairet: The `{}` look like an f-string.
```suggestion
url = multipass.generate_url(customer_data, "https://example.myshopify.com")
```

docs/multipass.md
```diff
@@ -0,0 +1,21 @@
+# Multipass
+[Shopify Multipass - Shopify Documentation](https://shopify.dev/docs/admin-api/rest/reference/plus/multipass) <br>
+This helper class generates token or URL that's needed for Shopify Multipass login (This feature is only available for Plus Store)
+```python
```
lishanl, line 5: I think it would be helpful to mention about enabling the multipass to obtain the secret. And that the URL generated with valid token is for redirecting the customer to the shopify store with and logged into the store.

hillairet (changes requested):
@lishanl @ponty33 Let's talk about the design.

Is there a reason to have a class?

I would prefer a set of functions instead:
```python
def generate_token(secret: str, contact_raw_data: Dict[str, Any]) -> bytes:
￼   ...
￼
def generate_url(secret: str, contact_raw_data: Dict[str, Any], url: str) -> str:
    ...
```
The docs could show how to use partial to avoid passing around the secret.
The user can still have `multipass.`:
```python
from spylib import multipass

url = multipass.generate_url(...
```
The small helper functions should start with `_` (that applies to internal methods too):
```python
def _get_keys(secret: str) -> Tuple[str, str]:
    key = SHA256.new(secret.encode('utf-8')).digest()
￼   return (key[0:16], key[16:32])
```
Bonus point if you use a dataclass or a pydantic model instead of a Tuple!  :wink:

lishanl:
> @lishanl @ponty33 Let's talk about the design.
> 
> Is there a reason to have a class?
> 
> I would prefer a set of functions instead:
> 
> ```python
> def generate_token(secret: str, contact_raw_data: Dict[str, Any]) -> bytes:
> ￼   ...
> ￼
> def generate_url(secret: str, contact_raw_data: Dict[str, Any], url: str) -> str:
>     ...
> ```
> 
> The docs could show how to use partial to avoid passing around the secret. The user can still have `multipass.`:
> 
> ```python
> from spylib import multipass
> 
> url = multipass.generate_url(...
> ```
> 
> The small helper functions should start with `_` (that applies to internal methods too):
> 
> ```python
> def _get_keys(secret: str) -> Tuple[str, str]:
>     key = SHA256.new(secret.encode('utf-8')).digest()
> ￼   return (key[0:16], key[16:32])
> ```
> 
> Bonus point if you use a dataclass or a pydantic model instead of a Tuple! 😉

OOP VS Functional programming 🥁

hillairet:
@lishanl Well maybe ... is "multipass" really an object though?
I don't think having a bunch of class means doing OOP, if none of your classes actually represent objects.

ponty33:
There wasn't much reason behind it. The code was referenced from an unmaintained repo, and it was written in class

I don't have much opinion on it, either way is fine to me

docs/multipass.md
ponty33: I don't really know the use of it. If it has no other use I can exclude it from doc

docs/multipass.md
ponty33: That is all in the Shopify doc. I think refer it to the official doc is better

hillairet: I'm with @ponty33 that it's best to let the Shopify docs speaks for itself and not duplicate. This way if it changes, we don't become out of date.

docs/multipass.md
hillairet: Since it's done and it doesn't hurt to expose it, let's keep it that way. Maybe we can't think of a scenario but it exists nonetheless.

spylib/multipass.py
```diff
@@ -0,0 +1,38 @@
...
+        iv = get_random_bytes(AES.block_size)
+        cipher = AES.new(self.encryption_key, AES.MODE_CBC, iv)
+        return iv + cipher.encrypt(str.encode(plainText))
+
+    def sign(self, secret):
+        return HMAC.new(self.signature_key, secret, SHA256).digest()
+
+    def pad(self, s):
+        return s + (AES.block_size - len(s) % AES.block_size) * chr(
+            AES.block_size - len(s) % AES.block_size
```
hillairet: Definitely prefix these function names with `_` to show that they are internal.

hillairet (changes requested):
@ponty33 Let's go with functions here please. No reason for a class.

docs/multipass.md
hillairet: @ponty33 Maybe let's have a sentence "If for some reason you need it, you can also generate separately the token used in the URL:" just above the `generate_token` example in the docs.

spylib/multipass.py
```diff
@@ -0,0 +1,40 @@
...
+import json
+from base64 import urlsafe_b64encode
+from typing import Any, Dict
+
+from Crypto.Cipher import AES
+from Crypto.Hash import HMAC, SHA256
+from Crypto.Random import get_random_bytes
+
+
+def generate_token(secret: str, contact_data: Dict[str, Any]) -> bytes:
```
hillairet: Multipass is to connect to a customer account so let's make this variable name more meaningful:
```suggestion
def generate_token(secret: str, customer_data: Dict[str, Any]) -> bytes:
```

spylib/multipass.py
```diff
@@ -0,0 +1,40 @@
...
+    key = SHA256.new(secret.encode('utf-8')).digest()
+    encryption_key = key[0:16]
+    signature_key = key[16:32]
+
+    contact_data['created_at'] = datetime.datetime.utcnow().isoformat()
+    cypher_text = _encrypt(encryption_key, json.dumps(contact_data))
+    return urlsafe_b64encode(cypher_text + _sign(signature_key, cypher_text))
+
+
+def generate_url(secret: str, contact_data: Dict[str, Any], url) -> str:
```
hillairet:
```suggestion
def generate_url(secret: str, customer_data: Dict[str, Any], store_url) -> str:
```

docs/multipass.md
```diff
@@ -0,0 +1,19 @@
...
+
+```python
+from spylib import multipass
+
+
+# Customer email is required to generate token or URL
+contact_data = {"email": "customer@email.com"}
+
+# Generate URL
+url = multipass.generate_url("MULTIPASS_SECRET", contact_data, 'https://example.myshopify.com')
```
hillairet: Having the arguments named makes the example even more readable:
```suggestion
url = multipass.generate_url(secrect="MULTIPASS_SECRET", customer_data=customer_data, store_url='https://example.myshopify.com')
```

docs/multipass.md
```diff
@@ -0,0 +1,19 @@
...
+# https://example.myshopify.com/account/login/multipass/{MultipassToken}
+
+# If for some reason you need the token, you can also generate the token used in the URL separately:
+token = multipass.generate_token("MULTIPASS_SECRET", contact_data)
```
hillairet: Here too with named arguments

hillairet (changes requested):
I believe the email is a required field and the only required field in the customer data dict. Please add a check so that SPyLib fails with a meaningful error message before the user makes the multipass API call.

docs/multipass.md
```diff
@@ -0,0 +1,23 @@
...
+# Generate URL
+url = multipass.generate_url(
+    secret="MULTIPASS_SECRET",
+    customer_data=customer_data,
+    store_url='https://example.myshopify.com'
+)
+# https://example.myshopify.com/account/login/multipass/{MultipassToken}
+
+# If for some reason you need the token, you can also generate the token used in the URL separately:
+token = multipass.generate_token(secret="MULTIPASS_SECRET", customer_data=customer_data)
```
lishanl: We can have single quotes for the strings to be consistent

spylib/multipass.py
```diff
@@ -0,0 +1,48 @@
...
+    cypher_text = _encrypt(encryption_key, json.dumps(customer_data))
+    return urlsafe_b64encode(cypher_text + _sign(signature_key, cypher_text))
+
+
+def generate_url(secret: str, customer_data: Dict[str, Any], store_url) -> str:
+    token = generate_token(secret, customer_data).decode('utf-8')
+    return f'{store_url}/account/login/multipass/{token}'
+
+
+def _encrypt(encryption_key, plainText) -> bytes:
```
lishanl: we can change `plainText` to `plain_text` to be consistent with the parameter casing

spylib/multipass.py
```diff
@@ -0,0 +1,48 @@
...
+    plainText = _pad(plainText)
+    iv = get_random_bytes(AES.block_size)
+    cipher = AES.new(encryption_key, AES.MODE_CBC, iv)
+    return iv + cipher.encrypt(str.encode(plainText))
+
+
+def _sign(signature_key, secret):
```
lishanl: I personally got confused by the `secret`, I thought it referred to the multipass secret. Maybe rename it to `cypher_text` to avoid the confusion?

lishanl (approved):
I am picking on the consistency here. 😅 Hope you don't mind Frank. It would be nice to have type annotations for all functions.

docs/multipass.md
ponty33: Sure, I am just too used to double quote :sweat_smile:

spylib/multipass.py
ponty33: Yeah, I think I copy & pasted so it's in camel case

spylib/multipass.py
hillairet: Maybe `secret` is not quite clear but it is what it's called in Shopify's documentation and UI:
image

Therefore we need to have this word. How about `multipass_secret` to make it very explicit?

spylib/multipass.py
```diff
@@ -0,0 +1,48 @@
...
+from Crypto.Cipher import AES
+from Crypto.Hash import HMAC, SHA256
+from Crypto.Random import get_random_bytes
+from pydantic import BaseModel
+
+
+class CustomerData(BaseModel):
+    email: str
+
+
```
hillairet: Not used, right?
```suggestion
```

spylib/multipass.py
hillairet: Oh I misread! My bad! You didn't change the `secret` I thought!  :wink:

hillairet (changes requested):
Remove unused thing. Otherwise looks great.

hillairet (approved):

ponty33:
Finally closed after 329 days :smile:

lishanl:
Definitely top 1 on the leaderboard for the longest open PR that actually gets completed at the end. 🏆 
Not to encourage this. But yea. Glad it's completed. 😆

hillairet:
I miss PullPanda for that! The listing of records and averages was great!

PR #10, opened 2021-07-22
Use callback_path variable in the redirect uri

lishanl:
Right now the oauth `redirect_uri` has the callback path hard coded. It doesn't work when the `callback_path` is set to something else. So here replace the hardcoded callback path with the `callback_path` variable


hillairet (approved):
Good catch!

PR #13, opened 2021-07-29
Add operation name for GraphQL

### LINKED ISSUE #12: Add operationName argument to the GraphQL calls
In one "query" string, one can pass [multiple operations](https://dgraph.io/docs/graphql/api/multiples/#multiple-operations) but then to run one of them, one has to specify the `operationName` field.

Let's add this as an optional parameter to the graphql call so we can write GraphQL files with multiple operations in them. It's very handy.

Nathan-Nesbitt:
This introduces the operation name functionality for GQL queries.

### DIFF
--- a/spylib/store.py
+++ b/spylib/store.py
@@ -155,13 +155,7 @@ class Store:
             'Content-type': 'application/json',
             'X-Shopify-Access-Token': self.access_token,
         }
-        body = {
-            'query': query,
-            'variables': variables,
-        }
-
-        if operation_name:
-            body['operationName'] = operation_name
+        body = {'query': query, 'variables': variables, 'operationName': operation_name}
 
         resp = await self.client.post(
             url=url,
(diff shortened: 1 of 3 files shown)


spylib/store.py
```diff
@@ -138,16 +145,28 @@ async def shoprequest(self, goodstatus, debug, endpoint, **kwargs) -> Dict[str,
...
             'Content-type': 'application/json',
             'X-Shopify-Access-Token': self.access_token,
         }
+        body = {
+            'query': query,
+            'variables': variables,
+        }
+
+        if operation_name:
+            body['operationName'] = operation_name
```
hillairet: I _think_ that works:
```suggestion
            'variables': variables,
            'operationName': operation_name,
        }
```

spylib/store.py
```diff
@@ -167,6 +186,17 @@ async def execute_gql(self, query: str, variables: Dict[str, Any] = {}) -> Dict[
                 raise ShopifyThrottledError(
                     f'Store {self.name}: The Shopify API token is throttling. '
                 )
+            if OPERATION_NAME_REQUIRED_ERROR_MESSAGE in errorlist:
+                raise ShopifyCallInvalidError(
+                    f'Store {self.name}: Operation name was required for this query. \
+                        This likely means you have multiple queries within one call \
+                        and you must specify which to run.'
+                )
```
hillairet: Very nice message!
You can actually do that because it's all surrounded by brackets. I think it's nicer.
```suggestion
                raise ShopifyCallInvalidError(
                    f'Store {self.name}: Operation name was required for this query. '
                    'This likely means you have multiple queries within one call '
                    'and you must specify which to run.'
                )
```

hillairet (approved):
Cool. Good to have. Nice error handling.
Let's merge that and next week let's rewrite your test together is a more organized and maintainable way.

spylib/store.py
Nathan-Nesbitt: It should work, I was worried about it throwing this error which happens when you pass in `operationName=Null`:
image
But I looked closer and it only happens if you have more than 1 named query.

hillairet (approved):

PR #14, opened 2021-08-03
Credit rate limit for GraphQL

### LINKED ISSUE #11: Handle credit rate limit in GraphQL code
We have the full rate limit handling setup for Rest but not for GraphQL. For the latter we just have retry for now.
We need to handle the rate limit in GraphQL too.

Nathan-Nesbitt:
Added in the logic for handling GraphQL rate limiting. It currently calculates the minimum amount of time required to be able to re-fill the bucket before attempting to re-submit the query. It completely kills the request if it will never run (> 1000).

### DIFF
--- a/spylib/store.py
+++ b/spylib/store.py
@@ -179,8 +179,10 @@ class Store:
             )
             if MAX_COST_EXCEEDED_ERROR_CODE in error_code_list:
                 raise ShopifyExceedingMaxCostError(
-                    f'Store {self.name}: The Shopify API token is throttling.'
-                    f' Query cost is too large (>{GRAPH_MAX}) and will never run'
+                    f'Store {self.name}: This query was rejected by the Shopify'
+                    f' API, and will never run as written, as the query cost'
+                    f' is larger than the max possible query size (>{GRAPH_MAX})'
+                    ' for Shopify.'
                 )
             elif THROTTLED_ERROR_CODE in error_code_list:  # This should be the last condition
                 query_cost = jsondata['extensions']['cost']['requestedQueryCost']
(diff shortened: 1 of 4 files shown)


spylib/store.py
```diff
@@ -163,10 +170,25 @@ async def execute_gql(self, query: str, variables: Dict[str, Any] = {}) -> Dict[
...
+            error_code_list = '\n'.join(
+                [
+                    err['extensions']['code']
+                    for err in jsondata['errors']
+                    if 'extensions' in err and 'code' in err['extensions']
+                ]
+            )
+            if MAX_COST_EXCEEDED_ERROR_CODE in error_code_list:
+                raise ShopifyExceedingMaxCostError(
+                    f'Store {self.name}: The Shopify API token is throttling.'
```
hillairet: Question of English: Is it really "throttling" when Shopify will just plainly refuse the query cause it's too big???

hillairet (approved):
Good for now. Let's have these tests for our test review tomorrow.
Also this is a good first pass, a nice improvement, but it's not ideal if we have a bunch of calls that happen at the same time (think a 100).
The GraphQL rate limiting should be similar to the rest stuff. Could be done in another PR, now that you have the tests.

spylib/store.py
Nathan-Nesbitt: Good question, I'd say you're right because even if the pool was full the query would never run!

Nathan-Nesbitt:
As the three tests were significantly different from each other (1 error, 1 success, 1 with multiple repeated calls to success), I changed the tests to not take params and instead be three separate functions.

This should address the issues from before, in a separate PR I will address the rest of #15 @hillairet as this should probably be split into separate files.

PR #17, opened 2021-08-11
Altered test folder structure

### LINKED ISSUE #15: Improve store tests organization
* Split GraphQL and Rest into two files
* Find out if `pytest.param` also takes named arguments and if so use named arguments
* Split into happy path and failures
* Get rid of side_effects function and use parametrization instead
* Have `base_gql_response` and in parameters have `mod_gql_response` so that the parametrization is minimal

Nathan-Nesbitt:
This provides a potentially more maintainable structure for the tests in spylib. 

The following is the core structure:

image

The tests are first broken down based off the main file they are stored in (e.g. store or oauth), then they are split on the methods for each class (e.g. store has the rest and graphql methods).

Finally, to allow for <250 line test files for easy reading, tests are split into individual files based on the core tests that were written.  

At the moment I have left the naming of the original test functions, for example `test_store_graphql_operation_name_happypath`. But by using this folder structure, these test names can be used to store more information about the individual test without overwhelming the reader. 

For example the previous test could be renamed to `test_happypath` for simple reading. This would change the output from the test from this:

```
tests/store/graphql/test_operation_name.py::test_store_graphql_operation_name_happypath PASSED                                                                                                                
```

To this:

```
tests/store/graphql/test_operation_name.py::test_happypath PASSED                                                                                                                
```


hillairet (approved):
Cool.
For rest and store you could have just `/tests/store/rest.py` and `/tests/oauth.py`. Then we break them down later on when/if needed. Just a suggestion. What you did works well.

PR #19, opened 2021-08-25
Refactor store

Nathan-Nesbitt:
This refactors the store so it can accept online and offline tokens. It provides some small changes to the structure of the class and some associated methods.


Nathan-Nesbitt:
This is still being worked on, I made the PR as I wanted to add comments as I go!

Nathan-Nesbitt:
I wrote out the solution as described, but there is an issue isolating the tests. I am currently checking to be sure a store, using an offline token, does not have the attribute `staff_id`. If I run the tests individually they pass, but if I run the full unit tests I get the following error:

```python
    @pytest.mark.asyncio
    async def test_offline_happypath():
        """
        This checks to see if the Store object has been initialized with the proper
        information for an offline token.
        """
        # We first save the token
        TestStore.save_offline_token(store_name='test-store', key='123')
    
        store = TestStore.load(store_name='test-store')
    
        assert store.store_name == 'test-store'
        assert store.access_token == '123'
>       assert not hasattr(store, 'staff_id')
E       AssertionError: assert not True
E        +  where True = hasattr(<store.shared.TestStore object at 0x7f5c644b3940>, 'staff_id')

tests/store/token/test_base.py:36: AssertionError
```

I think this is because the stores are stored globally (as a static class variable). What do you suggest @hillairet ?

hillairet:
So you didn't go for the `OnlineToken` and `OfflineToken` classes?
I didn't see the use of store and application classes on top but the separation into the two different classes actually makes sense!
It's awkward to have "save_online_token" and "save_offline_token". That's a sign that this store class has too much (and yes that's my design). We should just have the two classes as you suggested and then for each "save_token".
We have ST PIM that uses only the offline token of a store. I can imagine cases where an app would use only the online token ...

For your question about the `staff_id` attribute, well it'll go away if you split into two classes, but also I'd test for `assert store.staff_id == None` and make sure that the `staff_id` attribute is always set to something. I think it's cleaner and safer as you may have demonstrated.

PR #21, opened 2021-09-03
Refactoring the store to support token objects

Nathan-Nesbitt:
This went beyond a small change so I have moved this to a new branch from #19


Nathan-Nesbitt:
I thought I should explain this PR because in reality it should have been multiple PRs instead of one big PR (my apologies). I  realized after I made the changes how it could have been broken down, but retroactively changing this into multiple PRs would be crazy complicated and costly to test. 

The main goal was to fix the tokens to support both online and offline tokens. Because the store was using static class variables to store the `Store` instances it became a hassle to test the tokens. The `Store` was separated into an Application class which held a set of `Store` objects. Each store could then hold a single offline token, along with an online token for each admin user in the store.

SpyLib

As the previous tests for OAuth were written to only support offline tokens they needed to be re-implemented to ensure that the token class was working properly. I was trying to figure out how the functions were working and was running into issues determining what variables were used in what order while I was working on the Token/OAuth tests. For example, I was trying to follow the `jwt_key` value and its origin:

https://github.com/SatelCreative/satel-spylib/blob/3bfc32f5eee8d632d85ebd5c54e37e3d9966d4a6/spylib/oauth/redirects.py#L16

Side note: This `jwt_key` variable is actually the client secret. I know why this happened, because shopify randomly interchanged their naming throughout the original document. I changed it to just be client_secret because it is easier to follow.

To fix this realized that we could just pull all of the required information from the newly created `Application` class instead of redefining all of the variables for each of the routers. I merged these functions into the `Application` which made it easier to follow what the meaning of the OAuth variables were, It also removed config files like the following:

https://github.com/SatelCreative/satel-spylib/blob/3bfc32f5eee8d632d85ebd5c54e37e3d9966d4a6/spylib/oauth/config.py#L4-L13

This change was largely because you can't back-reference variables in the git repo. This feature is really helpful when you are learning how a repo works and helps with debugging. For example, you should now be able to follow most of the classes to how they are being used in the repo using the web interface:
image

A large portion of this PR is tests, they give 89% coverage. I can add more to test the edge cases but I am going to hold off to try to reduce the size.

hillairet:
So I am quite confused about the application and store classes as I thought [we came to an agreement that they were not needed](https://satel.slack.com/archives/G017U0W6LR3/p1629763369050500?thread_ts=1629762591.049300&cid=G017U0W6LR3). The OnlineToken and OfflineToken are definitely useful as [I mentioned before](https://github.com/SatelCreative/satel-spylib/pull/19#issuecomment-905873163) but the Store and Application don't add any feature that we need *right now*. Maybe in the future we'll need them but we should be for what we need and not what we think is nice or might be needed in the future maybe.

PR #23, opened 2021-09-09
Feature/token refactor

### LINKED ISSUE #18: Refactor Store class
# What, Why

To handle the online token #9 we need to rethink how we are doing the whole Store class.
At the same time let's improve the API with an overall rewrite.

# How

1. StoreAPI or StoreABC (Naming is hard) abstract class with methods that need to be overloaded:
    * `save_online_token` and `save_offline_token` to be called at the end of the OAuth process to store in DB, or wherever needed/wanted
    * `load_online_token` and `load_offline_token` to be called whenever the `load` method initiates a store
2. `load` function to instantiate the Store class with an async call:
        * If argument is just store name (store subdomain) then load offline token: `store = await Store.load(storename=storename)`
        * If argument is the store name and the staff id then load online token: `store = await Store.load(storename=storename, staff_id=staff_id)`
3. Rename `shoprequest` to something `execute_rest`

Nathan-Nesbitt:


spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+        )
+
+
+class OfflineTokenABC(Token, ABC):
+    """
+    Offline tokens are used for long term access, and do not have a set expiry.
+    """
+
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
```
hillairet: Not needed
```suggestion
```

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+
+    async def reset_token(self, client_id: str, client_secret: str, code: str):
+
+        token: OfflineTokenResponse = OfflineTokenResponse(
+            **await self._get_token(client_id, client_secret, code)
+        )
+        self.access_token = token.access_token
+        self.scope = token.scope.split(',')
```
hillairet: Let's keep this first pass simpler and let's not handle the token reset for now. Since it will involve the oauth process, we need this to be done properly so let's do it later.
```suggestion
```

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+
+
+@dataclass
```
hillairet: Why is this one a dataclass???

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+        client_secret: str,
+        code: str,
+    ):
+        """
+        Creates a new instance of the object and requests the token from the API
+        as it requires an async call.
+        """
+        token = cls(store_name=store_name)
+        await token.reset_token(client_id, client_secret, code)
+        return token
```
hillairet: Same comment, this is part of the oauth function for now, let's keep it there and maybe move it here later on, if it makes sense.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+    should either be using the OfflineToken or the OnlineToken.
+    """
+
+    store_name: str
+    scope: List[str]
+    access_token: Optional[str]
+    access_token_invalid: bool
+    client: AsyncClient
+    oauth_url: str
+    api_url: str
```
hillairet: These are generated from the `store_name` and the `api_version` so don't make them static and define them as `@property` methods.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+    """
+    Abstract class for token objects. This should never be extended, as you
+    should either be using the OfflineToken or the OnlineToken.
+    """
+
+    store_name: str
+    scope: List[str]
+    access_token: Optional[str]
+    access_token_invalid: bool
+    client: AsyncClient
```
hillairet: We want SPyLib to have its own client, not have to receive it from the user
```suggestion
```

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+    oauth_url: str
+    api_url: str
+    api_version: str = '2021-04'
```
hillairet: Interesting idea to hardcode it but allow the user to overwrite. Indeed in theory at this point SPyLib should be version agnostic!
But it's not something that should be modifiable for each instance of the class so use that instead:
```suggestion
    api_version: ClassVar[str] = '2021-04'
```
`ClassVar` comes from `typing`
Now it can be overwritten in the inherited class but not when declaring each instance.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+    rest_bucket_max: int = 80
+    rest_leak_rate: int = 4
+    graphql_bucket_max: int = 1000
+    graphql_leak_rate: int = 50
```
hillairet: `ClassVar` here too. No point in the user setting these values. We need defaults but at the first API call the response header will give us the proper values. Much better way to do it.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+
+class Token(ABC, BaseModel):
+    """
+    Abstract class for token objects. This should never be extended, as you
+    should either be using the OfflineToken or the OnlineToken.
+    """
+
+    store_name: str
+    scope: List[str]
+    access_token: Optional[str]
```
hillairet, line 63: Might be due to having some oauth stuff but I don't think this should be Optional.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+        await token.reset_token(client_id, client_secret, code)
+        return token
+
+    @abstractmethod
+    async def save_token(self):
+        """
+        This method handles saving the token. By default this does nothing,
+        therefore the developer should override this.
+        """
+        raise NotImplementedError("Please Implement the save_token method")
```
hillairet: It's an abstract method so you don't need to do this manually. The class will do it for you if this method is not overwritten.
```suggestion
        pass

File: spylib/token.py [hunk 11]
```diff
@@ -0,0 +1,428 @@
...
+        return response.json()
+
+    async def reset_token(self, client_id: str, client_secret: str, code: str):
+        """Use this function to initialize a new access token for this store"""
+        await self._get_token(client_id, client_secret, code)
+        self.access_token_invalid = False
+
+    @classmethod
+    def get_unique_id(cls) -> str:
+        return ShortUUID().random(length=10)
```
hillairet: Actually this should come from Toolip. This is a pure function. Nothing to do with this class.
Also a classmethod should return the class it belongs too, though not actually enforced. I've used them to return lists of the parent class ... that may not be pythonic but I did it anyway  :wink:

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+            self.__add_new_tokens()
+        self.rest_bucket -= 1
+
+    def __add_new_tokens(self):
+        now = monotonic()
+        time_since_update = now - self.updated_at
+        new_tokens = floor(time_since_update * self.rate)
+        if new_tokens > 1:
+            self.rest_bucket = min(self.rest_bucket + new_tokens, self.rest_bucket_max)
+            self.updated_at = now
```
hillairet: I propose to rename these "tokens" to something like `credits` or maybe `rest_credits` to make the code less confusing. This name always bothered me. Good time to fix this.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+        raise NotImplementedError("Please Implement the save_token method")
+
+    @classmethod
+    @abstractmethod
+    def load_token(self):
+        """
+        This method handles loading the token. By default this does nothing,
+        therefore the developer should override this.
+        """
+        raise NotImplementedError("Please Implement the load_token method.")
```
hillairet: Same
```suggestion
        pass
```

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+
+    async def _get_token(
```
hillairet: A better name would be `_get_from_shopify` but let's not implement that (yet?).
Step by step. Incremental and iterative development.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+    def __new__(cls, *args, **kwargs):
+        """
+        Function to keep user from instantiating an instance of this class.
+        """
+        if cls is Token:
+            raise TypeError(f'Can\'t instantiate abstract class {cls.__name__} directly')
+        return object.__new__(cls)
+
+    def save_json(self):
+        json.dumps(
```
hillairet: What's that for?
pydantic has a method for that already.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+        super().__init__(
+            **dict(
+                **kwargs,
+                **{
+                    'associated_user_scope': associated_user_scope,
+                    'associated_user': associated_user,
+                    'expires_at': datetime.now() + timedelta(days=0, seconds=expires_in),
+                },
+            )
+        )
```
hillairet: You shouldn't have to define `__init__`. That's one thing I really like with dataclasses and better in pydantic `BaseModel`

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+@dataclass
+class OnlineTokenABC(Token):
+    """
+    Online tokens are used to implement applications that are authenticated with
+    a specific user's credentials. These extend on the original token, adding
+    in a user, its scope and an expiry time.
+    """
+
+    associated_user_scope: Optional[List[str]]
+    associated_user: Optional[AssociatedUser]
```
hillairet: Do we really need to have all the user's info? Seems like the Shopify ID would be enough.

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+
+@dataclass
+class OnlineTokenABC(Token):
+    """
+    Online tokens are used to implement applications that are authenticated with
+    a specific user's credentials. These extend on the original token, adding
+    in a user, its scope and an expiry time.
+    """
+
+    associated_user_scope: Optional[List[str]]
```
hillairet: You already have `scope` in the `Token` class.
```suggestion
```

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+        """
+        pass
+
+    @classmethod
+    def load_token(cls, store_name: str, associated_user: str):
+        """
+        This method handles loading the token. By default this does nothing,
+        therefore the developer should override this.
+        """
+        pass
```
hillairet: Wait I just realized that you shouldn't have `load_token` in `Token` because the method arguments depends if you are in `OfflineToken` or `OnlineToken`.

Also this has to be async
```suggestion
    @classmethod
    async def load_token(cls, store_name: str, associated_user: str):
        """
        This method handles loading the token. By default this does nothing,
        therefore the developer should override this.
        """
        pass
```

spylib/token.py
```diff
@@ -0,0 +1,428 @@
...
+        self.associated_user = token.associated_user
+        self.associated_user_scope = token.associated_user_scope.split(',')
+        self.expires_at = datetime.now() + timedelta(days=0, seconds=token.expires_in)
+
+    async def save_token(self):
+        """
+        This method handles saving the token. By default this does nothing,
+        therefore the developer should override this.
+        """
+        pass
```
hillairet: From `Token`
```suggestion
```

hillairet (changes requested):
I know it's a work in progress so probably some of the comments are not applicable. But I thought I'd check things out to guarantee alignment early on.

spylib/token.py
Nathan-Nesbitt: I was reading the docs and they say that the user can have a subset of the app scopes for online tokens, which is returned in the token response. This is what the OAuth endpoint returns:

https://github.com/SatelCreative/satel-spylib/blob/e4375f8a6a072fec2f36a4631b0cf7389806ed9a/spylib/token.py#L52-L57


Should I leave it?

spylib/token.py
Nathan-Nesbitt: Yeah totally, I just wrote it to have a list of all possible data so we know what is being returned from the endpoint. I can change it so we only keep ID.

spylib/token.py
Nathan-Nesbitt: Smart I like it

spylib/token.py
Nathan-Nesbitt: Should I keep it aligned with the shopify docs or come up with our own names?  I think they refer to them as buckets and the leak for both graphql and rest. But if you have a different name that would be better we can go with that

Nathan-Nesbitt:
@hillairet This looks like a big PR, most of the changes are just moving files around in testing to support the tests for the two new tokens. 

This is what I have at the moment, these were just quick tests to be sure that what I changed works, but I think it could just be one unit test with different parameters for the online and offline tokens:
image

Let me know your preference for layout!

hillairet:
Cool looks like we are getting there.
We have to be careful not to retest the same code and double our tests code base.
The graphql and rest methods belong to `Token`, not `OfflineTokenABC` or `OnlineTokenABC`. So there is no point testing the GraphQL and Rest code twice. You can create a class in your test code inheriting from Token and testing Rest and GraphQL directly. We don't need to test python's abstract classes mechanisms.
Testing the initialization is fine though with abstract classes, not sure what we are really testing.
Also let's not make the directory structure more complicated and deep than it needs to be:
```
/tests
     |- token
            |- test_initialization.py    # for the initialization of OfflineToken and OnlineToken
            |- test_graphql_base.py
            |- test_graphql_operation_name.py
            |- test_graphql_throttling.py
            |- test_rest.py
```

Also instead of `shared.py` you should use pytest fixtures which will be much better than having `from ...shared import xxx`

spylib/token.py
Nathan-Nesbitt: Based on your previous comment I also removed load_token from the Token class.

spylib/token.py
Nathan-Nesbitt: This was because the _get_token method was written as a method  instead of a class method requiring a bunch of parameters. 

https://github.com/SatelCreative/satel-spylib/blob/6f1d9f1d16aa0f8a47236ec7e02ba5dba30cc680/spylib/token.py#L150-L158

I will remove this from this PR based on your previous suggestions. Would you like me to open an issue to discuss the integration of the OAuth features? As I have some other thoughts on the process that might make it easier.

spylib/token.py
Nathan-Nesbitt: I can change this, but will there ever be a scenario where we have an application that needs to support both regular and Shopify plus? Because making it a class variable will remove the ability to support that scenario.

spylib/token.py
Nathan-Nesbitt: This has been changed to `__await_rest_bucket_refill` and `__fill_rest_bucket` in the latest changes sticking with shopify's naming scheme. Let me know if this is okay!

Nathan-Nesbitt:
This should be stable now, there is one error in the typing 

```
spylib/token.py:235: error: Argument "headers" to "post" of "AsyncClient" has incompatible type "Dict[str, Optional[str]]"; expected "Union[Headers, Dict[str, str], Dict[bytes, bytes], Sequence[Tuple[str, str]], Sequence[Tuple[bytes, bytes]], None]"
```

This comes from the `AsyncClient.post()` method, and doesn't match their docs for what the expected input is.

I have opened a discussion with the devs on httpx as it is not clear what the input should be, but I will update this as soon as I know what's going wrong.

spylib/token.py
hillairet: Oh I didn't realize that Shopify sends the app scope AND the user scope for online tokens ... I thought it'd send only the online token one.
I don't see the point of `scope` here since what any online token will be able to do is what is allowed by the `associated_user_scope` ...
It also means that any code we write that uses the scope will have to be different for online and offline tokens. Sounds annoying. It seems to me it would be better to save `associated_user_scope` in our `scope`. Maybe we should rename the `scope` field to something like `token_scope` to clarify that it's not the app scope or the user scope, but the token scope so it depends on the type of token ...
We can use pydantic's aliases and have a different one for Online and Offline tokens. It's not pretty but that's what we have.

spylib/token.py
hillairet: Yeah let's deal with it in a later round in another issue

spylib/token.py
```diff
@@ -135,10 +152,10 @@ async def shoprequest(self, goodstatus, debug, endpoint, **kwargs) -> Dict[str,
                 jresp = response.json()
                 # Recalculate the rate to be sure we have the right one.
                 calllimit = response.headers['X-Shopify-Shop-Api-Call-Limit']
-                self.max_tokens = int(calllimit.split('/')[1])
+                self.rest_bucket_max = int(calllimit.split('/')[1])
                 # In Shopify the bucket is emptied after 20 seconds
                 # regardless of the bucket size.
-                self.rate = int(self.max_tokens / 20)
+                Token.rest_leak_rate = int(self.rest_bucket_max / 20)
```
hillairet: If it's a class variable then there will be one leak rate for all tokens. An app could be instead on a regular Shopify store and on a Shopify plus store, with two different leak rates.
```suggestion
                self.rest_leak_rate = int(self.rest_bucket_max / 20)
```

spylib/token.py
```diff
@@ -216,14 +238,42 @@ async def execute_gql(
...
+    """
+    Offline tokens are used for long term access, and do not have a set expiry.
+    """
+
+    @abstractmethod
+    async def save_token(self):
+        pass
+
+    @abstractclassmethod
+    def load_token(cls, store_name: str):
```
hillairet: Can we just call them `save` and `load` since they will be called on a token instance?
I'm lazy that way

spylib/token.py
```diff
@@ -216,14 +238,42 @@ async def execute_gql(
...
 
-class UniqueStore(Store):
-    __instance = None
+class OfflineTokenABC(Token, ABC):
+    """
+    Offline tokens are used for long term access, and do not have a set expiry.
+    """
+
+    @abstractmethod
+    async def save_token(self):
```
hillairet: Move that one to Token since it doesn't have arguments that depend on Online vs Offline token

spylib/token.py
hillairet: This method has less comments than the OnlineToken equivalent ... it's a little jealous ...  :wink:

spylib/token.py
hillairet: Yeah let's not force the user of the lib to have to load this data is they don't have to.

tests/token/conftest.py
```diff
@@ -0,0 +1,130 @@
...
+    test_database = {"offline": {}, "online": {}}
+
+
+@pytest.fixture(scope='session', autouse=True)
+def online_token_data() -> OnlineTokenResponse:
+    return OnlineTokenResponse(
+        access_token='ONLINETOKEN',
+        scope=','.join(['write_products', 'read_customers']),
+        expires_in=86399,
+        associated_user_scope=','.join(['write_orders', 'read_products']),
```
hillairet: As you read `associated_user_scope` has to be a subset of `scope` so:
```suggestion
        scope=','.join(['write_products', 'read_customers']),
        expires_in=86399,
        associated_user_scope=','.join(['read_products']),
```

tests/token/conftest.py
```diff
@@ -0,0 +1,130 @@
...
+
+@pytest.fixture(scope='session', autouse=True)
+def app_information() -> AppInformation:
+    return AppInformation()
+
+
+@pytest.fixture(scope='session', autouse=True)
+def MockHTTPResponse():
+    @dataclass
+    class MockHTTPResponse:
```
hillairet: Wait that works to have the two with the same name?
That's weird and confusing. Let's not do that please. Could be:
```suggestion
def MockHTTPResponse():
    @dataclass
    class MockHTTPResponseCls:
```

hillairet (changes requested):
The README.md needs a good update.

I know I suggested that but now looking at the initialization test, I'm not sure what they are testing. We want to be testing our code, not python itself. You wrote a save token and load token for the test at you are testing that ... not sure what we are verifying here ...
It's ok to have them but they are kind of useless, at least part of the test is.

tests/token/conftest.py
Nathan-Nesbitt: Good point, I think this should probably be checked in some middleware at some point?

hillairet: Hummmmm If you don't mind making an issue for us to get back to it. I have learned that you can't always prevent the users from doing something stupid. So while we want to prevent the preventable, we shouldn't over do it either

spylib/token.py
Nathan-Nesbitt: I think they do this because you are supposed to verify that the user scope that is returned is a subset of the app scope. You can re-request the auth for the user if the new scope of the application is less than the original scope for the user.

Should we add in the check as middleware so on every request using online tokens it checks to be sure that the token scopes are valid?

tests/token/test_rest.py
```diff
@@ -3,23 +3,32 @@
...
+    mocker,
+    offline_token_data: OfflineTokenResponse,
+    app_information: AppInformation,
+    MockHTTPResponse,
+):
+    token = Token(
+        store_name=app_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet: The token should be initialized like this only during the OAuth. Use `load` instead on an Offline or Online token (no need to do both). That's the point of defining the `test_database` in `conftest`, to test things the way they will be used by a user.
Same with the other Token initializations.

hillairet (changes requested):
Ok on closer second inspection:
* `test_token`, `test_online_token`, and `test_offline_token` are ok as they test some default values
* `test_save_token`, `test_load_token`, `test_save_offline_token` and `test_load_offline_token` test the testing code you wrote in `conftest.py` so let's remove them. Let's not maintain code testing the test code. When we have a large set of tests, we don't want to have to keep these ones up to date too.

spylib/token.py
hillairet: The validity of the scopes should be checked during the online token oauth process. That's enough. Once it was allowed and we stored it, it's good to go.

Ok so here is the way to go about it IMHO:
* Have a `token_scope` field that will have the given token's scope regardless if the token is offline or online
* Add `app_token` _later on if needed only_ to the OnlineToken. I wish Shopify had named them that way!

We haven't done this kind of scope comparison so far and I don't want to force the library user to record both the token scope and the app scope if we never use the latter for online tokens

spylib/exceptions.py
```diff
@@ -27,7 +27,7 @@ class ShopifyExceedingMaxCostError(ShopifyError):
     pass
 
 
-def not_our_fault(exception: Exception):
+def not_our_fault(exception: BaseException):
```
hillairet: We don't want to catch [system-exiting exceptions](https://docs.python.org/3/library/exceptions.html) so let's use `Exception`
```suggestion
def not_our_fault(exception: Exception):
```

spylib/token.py
```diff
@@ -29,63 +28,81 @@
...
-        self.rate = RATE
-
-    async def __wait_for_token(self):
-        self.__add_new_tokens()
-        while self.tokens <= 1:
+
+class Token(ABC, BaseModel):
+    """
+    Abstract class for token objects. This should never be extended, as you
+    should either be using the OfflineToken or the OnlineToken.
```
hillairet:
```suggestion
    should either be using the OfflineTokenABC or the OnlineTokenABC.
```

tests/token_classes.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @validator('headers', pre=True, always=True)
+    def set_id(cls, fld):
+        return fld or {'X-Shopify-Shop-Api-Call-Limit': '39/40'}
+
+    def json(self):
+        return self.jsondata
+
+
+class AppInformation(BaseModel):
```
hillairet: Why is that called "AppInformation" and not "StoreInformation"?

tests/token/test_rest.py
```diff
@@ -3,23 +3,27 @@
...
 from spylib.exceptions import ShopifyCallInvalidError
 
-from ..shared import MockHTTPResponse
+from ..token_classes import MockHTTPResponse, app_information, offline_token_data
 
 
 @pytest.mark.asyncio
 async def test_store_rest_happypath(mocker):
-    store = Store(store_id='TEST', name='test-store', access_token='Te5tM3')
+    token = Token(
```
hillairet: I'm confused ... Still not using `.load`   :persevere:

hillairet (changes requested):
More than I expected.
Now that there are changes ... In the docs you put online token before offline token. During the oauth we always get the offline token first since we get the app's token before we get a user's token. I'd follow the same order in the docs.

tests/token/test_rest.py
Nathan-Nesbitt: Oh because there is no load method on the base Token class. This shouldn't even happen because you aren't supposed to instantiate it as it's abstract. I did this based on a previous comment that we should be testing the functionality of the ABC using the ABC.

tests/token_classes.py
Nathan-Nesbitt: It (mostly) contains information specific to the application not the store. The `store_name` and `api_version` property is there because I needed somewhere to put a constant that could be shared across all tests. It could be put somewhere more descriptive if you'd like.

spylib/exceptions.py
Nathan-Nesbitt: Weird I was getting a type error yesterday because of this but now it's gone... Fixed on new commit

spylib/token.py
Nathan-Nesbitt: So it is safe to assume that the app scope will never change? I'm just thinking if the apps scope becomes more restrictive and removes an existing permission from an OnlineToken, there will be no way to know/renew this token. We will just have to dump and renew all tokens forcefully or users will run into weird errors they will have no control over.

tests/token/test_rest.py
hillairet: Ok that's where the misunderstanding is. I don't remember what I said but I didn't mean to test `Token` directly. You should use your `OfflineToken` or `OnlineToken` derived from `OfflineTokenABC and `OnlineTokenABC` respectively and use them as a user would. That will indirectly but in real conditions test `Token`.

tests/token_classes.py
hillairet: Ok so `TestInformation`?
Sorry the current name just really threw me off when I first saw it.

tests/token/test_rest.py
Nathan-Nesbitt: Ok great, so just nuke this test?

Nathan-Nesbitt:
@hillairet this is something I noticed when developing using this using the current design:

Screenshot from 2021-09-27 06-32-48

Since we aren't defining the `__init__` for the classes it is using the default text for a pydantic object. I know the hope was to not have to do this, but can we add in the `__init__` as it helps by showing the expected arguments? Unless you know a trick to get this functionality without it?

Nathan-Nesbitt:
Also should we include a fix for #34 in this? Or leave it for a different PR?

hillairet:
> @hillairet this is something I noticed when developing using this using the current design:
> 
> Screenshot from 2021-09-27 06-32-48
> 
> Since we aren't defining the `__init__` for the classes it is using the default text for a pydantic object. I know the hope was to not have to do this, but can we add in the `__init__` as it helps by showing the expected arguments? Unless you know a trick to get this functionality without it?

Please make an issue for this and let's do that later. This PR doesn't have to be perfect or complete.

tests/token/test_graphql_throttling.py
```diff
@@ -194,7 +202,11 @@ async def test_store_graphql_throttling_error_test(mocker):
     3. The query fails indefinitely due to it being in excess of the maximum
         possible query size.
     """
-    store = Store(store_id='TEST', name='test-store', access_token='Te5tM3')
+    token = Token(
+        store_name=test_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet:
```suggestion
    offline_token = await OfflineToken.load(store_name=test_information.store_name)
```

tests/token/test_graphql_throttling.py
```diff
@@ -109,7 +113,11 @@ async def test_store_graphql_throttling_catch_cap(mocker):
         the bucket to re-fill.
 
     """
-    store = Store(store_id='TEST', name='test-store', access_token='Te5tM3')
+    token = Token(
+        store_name=test_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet:
```suggestion
    offline_token = await OfflineToken.load(store_name=test_information.store_name)
```

tests/token/test_graphql_throttling.py
```diff
@@ -58,7 +58,11 @@ async def test_store_graphql_throttling_happypath(mocker):
 
     This handles the first case (#1).
     """
-    store = Store(store_id='TEST', name='test-store', access_token='Te5tM3')
+    token = Token(
+        store_name=test_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet:
```suggestion
    offline_token = await OfflineToken.load(store_name=test_information.store_name)
```

tests/token/test_graphql_operation_names.py
```diff
@@ -124,8 +140,12 @@ async def test_store_graphql_operation_name_badquery(query, operation_name, erro
...
     This tests the error cases.
 
-    '''
-    store = Store(store_id='TEST', name='test-store', access_token='Te5tM3')
+    """
+    token = Token(
+        store_name=test_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet:
```suggestion
    offline_token = await OfflineToken.load(store_name=test_information.store_name)
```

tests/token/test_graphql_operation_names.py
```diff
@@ -58,8 +64,12 @@ async def test_store_graphql_operation_name_happypath(
...
 
     This checks just the successful queries.
-    '''
-    store = Store(store_id='TEST', name='test-store', access_token='Te5tM3')
+    """
+    token = Token(
+        store_name=test_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet:
```suggestion
    offline_token = await OfflineToken.load(store_name=test_information.store_name)
```

tests/token/test_graphql_base.py
```diff
@@ -79,21 +89,25 @@ async def test_store_graphql_badquery(mocker):
...
 
 
 @pytest.mark.asyncio
 async def test_store_graphql_tokeninvalid(mocker):
-    store = Store(store_id='TEST', name='test-store', access_token='INVALID')
+    token = Token(
+        store_name=test_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet:
```suggestion
    offline_token = await OfflineToken.load(store_name=test_information.store_name)
```

tests/token/test_graphql_base.py
```diff
@@ -48,14 +52,20 @@ async def test_store_graphql_happypath(mocker):
 
 @pytest.mark.asyncio
 async def test_store_graphql_badquery(mocker):
-    store = Store(store_id='TEST', name='test-store', access_token='Te5tM3')
+    token = Token(
+        store_name=test_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet:
```suggestion
    offline_token = await OfflineToken.load(store_name=test_information.store_name)
```

tests/token/test_graphql_base.py
```diff
@@ -2,21 +2,25 @@
...
 
 
 @pytest.mark.asyncio
 async def test_store_graphql_happypath(mocker):
-    store = Store(store_id='TEST', name='test-store', access_token='Te5tM3')
+    token = Token(
+        store_name=test_information.store_name,
+        access_token=offline_token_data.access_token,
+        scope=offline_token_data.scope.split(','),
+    )
```
hillairet:
```suggestion
    offline_token = await OfflineToken.load(store_name=test_information.store_name)
```

spylib/token.py
hillairet: We currently have no process for the renewal anyway. We need to do that. That's another item that belongs in another issue/PR. Again we can add it later if needed but let's not over-engineer this.

hillairet (changes requested):
We might get there eventually.

hillairet (approved):
:tada: :partying_face: :confetti_ball:

PR #26, opened 2021-09-17
Feature/session tokens

### LINKED ISSUE #16: Implement the necessary backend piece for Shopify's session tokens
The [Shopify docs on the topic](https://shopify.dev/apps/auth/session-tokens/authenticate-an-embedded-app-using-session-tokens) seems pretty complete. We should implement it in SPyLib.

Nathan-Nesbitt:
Provides checking and typing for session tokens


spylib/utils/session_token.py
```diff
@@ -0,0 +1,112 @@
+from __future__ import annotations
+
+from typing import Any, ClassVar, Dict, Optional
+from pydantic.main import BaseModel
+from pydantic.networks import HttpUrl
+from datetime import datetime, timedelta
+
+import jwt
```
hillairet: Why not?
```suggestion
from jwt import decode
```

hillairet (changes requested):
Looks good.
Documentation needed, might as well have it in the same PR.
I'd suggest to the user to define a partial function:
```python
decode_session_token = partial(SessionToken.decode_token_from_header, api_key=api_key, secret=secret)
```
Then they can use that function in [FastAPI's `Depends`](https://fastapi.tiangolo.com/tutorial/dependencies/)
```python
@app.get("/items/")
async def read_items(session: SessionToken = Depends(decode_session_token)):
    return commons
```
I *THINK* that would work.
Got to be careful with partial and positional arguments ...

spylib/utils/session_token.py
Nathan-Nesbitt: I thought it would be clearer what decode does if I used it directly from `jwt`. Let me know if you have a preferred syntax.

README.md
```diff
@@ -195,4 +195,22 @@ result of the installation and the login processes respectivaly. They are meant
 particular to record the offline and online tokens in your app's database.
 They can be synchronous or asynchronous functions taking the storename and the token
 as arguments.
+
+### Session Tokens
+
+The session token functionality can be used to verify the session for the user.
+The suggested syntax is to define a partial function:
```
hillairet: Add link to Shopify's docs

spylib/utils/session_token.py
```diff
@@ -0,0 +1,122 @@
...
+        secret: str,
+    ) -> SessionToken:
+        # Verify the integrity of the token
+
+        # Take the authorization headers and unload them
+        token = cls.__extract_session_token(authorization_header)
+        payload = cls.__decode_session_token(token, api_key, secret)
+
+        # Verify enough fields specified and perform validation checks
+        session_token = cls(**payload)
```
hillairet: Read [this](https://github.com/orgs/SatelCreative/teams/python/discussions/43)
```suggestion
        session_token = cls.parse_obj(payload)
```

tests/utils/test_session_token.py
```diff
@@ -0,0 +1,111 @@
...
+
+from spylib.utils.session_token import (
+    InvalidIssuerError,
+    MismatchedHostError,
+    SessionToken,
+    TokenAuthenticationError,
+)
+
+api_key = "API_KEY"
+api_secret = "API_SECRET"
```
hillairet: Constants. See PEP8
```suggestion
API_KEY = "API_KEY"
API_SECRET = "API_SECRET"
```

tests/utils/test_session_token.py
```diff
@@ -0,0 +1,111 @@
...
+
+    assert session_token.iss == token['iss']
+    assert session_token.dest == token['dest']
+    assert session_token.aud == token['aud']
+    assert session_token.sub == token['sub']
+    assert session_token.exp == token['exp']
+    assert session_token.nbf == token['nbf']
+    assert session_token.iat == token['iat']
+    assert session_token.jti == token['jti']
+    assert session_token.sid == token['sid']
```
hillairet: pytest has a lovely dictionary comparison and error reporting so I'd use that instead:
```suggestion
    assert session_token.dict() == token
```

tests/utils/test_session_token.py
```diff
@@ -0,0 +1,111 @@
...
+    with pytest.raises(jwt.InvalidSignatureError):
+        SessionToken.decode_token_from_header(header, api_key, api_secret)
+
+
+@pytest.mark.asyncio
+async def test_invalid_hostname(token):
+    token['iss'] = 'https://someinvalidhost.com'
+    header = generate_auth_header(token)
+    with pytest.raises(InvalidIssuerError):
+        SessionToken.decode_token_from_header(header, api_key, api_secret)
```
hillairet: You should be able to parametrize this test such that you don't have to rewrite it 5 times ... copy-pasting is evil.
```suggestion
@pytest.mark.asyncio
async def test_invalid_hostname(token, fixture):
    token.update(fixture.token_mod)  # with token_mod = dict(iss='https://someinvalidhost.com')
    header = fixture.header  # My understanding is that token is always the same ... so this could be generated once as well, and could be '' if needed.
    with pytest.raises(fixture.expected_exception):
        SessionToken.decode_token_from_header(header, api_key, api_secret)
```

hillairet (changes requested):
Looks good overall. Improvement requests.

spylib/utils/session_token.py
```diff
@@ -0,0 +1,122 @@
...
+    class Config:
+        arbitrary_types_allowed = True
+
+    def validate_self(self) -> None:
+        self.__validate_hostname()
+        self.__validate_destination()
+        self.__validate_sub()
+
+    @classmethod
+    def decode_token_from_header(
```
hillairet: Since your class is `SessionToken`, you don't have to say `token` in the name here.
Plus we don't care if we effectively `decode`. It's nice to have a short `Session.from_header(...)` like pandas `DataFrame.read_csv`
```suggestion
    def from_header(
```

spylib/utils/session_token.py
```diff
@@ -0,0 +1,122 @@
...
+    iat: Optional[float]
+    jti: int
+    sid: str
+
+    algorithm: ClassVar[str] = "HS256"
+    prefix: ClassVar[str] = "Bearer "
+    required_fields: ClassVar[List[str]] = ["iss", "dest", "sub", "jti", "sid"]
+
+    class Config:
+        arbitrary_types_allowed = True
```
hillairet: Why using that???

spylib/utils/session_token.py
```diff
@@ -0,0 +1,122 @@
...
+    sub: int
+    exp: Optional[float]
+    nbf: Optional[float]
+    iat: Optional[float]
+    jti: int
+    sid: str
+
+    algorithm: ClassVar[str] = "HS256"
+    prefix: ClassVar[str] = "Bearer "
+    required_fields: ClassVar[List[str]] = ["iss", "dest", "sub", "jti", "sid"]
```
hillairet: Do we really want to make that customizable?
You could use the `@property` trick to make them class constants and organized.

spylib/utils/session_token.py
```diff
@@ -0,0 +1,122 @@
...
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    def validate_self(self) -> None:
+        self.__validate_hostname()
+        self.__validate_destination()
+        self.__validate_sub()
```
hillairet: Why not use [pydantic validators](https://pydantic-docs.helpmanual.io/usage/validators/)?

hillairet (changes requested):
I missed that on the first pass, that's on me: The SessionToken class is overly complicated.
`__url_base` makes sense as a method because you call it about 5 times, but `__validate_*` and `__decode_session_token` are called once and are super small. It's ok for `from_header` to have it all. With empty lines and comments it'll be super clear and won't require any jumping around when reading the code.
Python zen: Readability counts!

spylib/utils/session_token.py
Nathan-Nesbitt: Good point, this was how the official library originally implemented them. I will reformat!

spylib/utils/session_token.py
Nathan-Nesbitt: Good point, will fix.

spylib/utils/session_token.py
Nathan-Nesbitt: Done!

spylib/utils/session_token.py
Nathan-Nesbitt: Actually bit of a problem, `@property` isn't itterable. So  `required_fields` is going to be a problem as it will require a custom class to implement the itterable property. 

I also have to cast every call to the `prefix` property back to string as all of the functions are expecting values of type `str` not `property` type.

I will wait and see if you think this is worth implementing or if it is over-complicating it.

spylib/utils/session_token.py
```diff
@@ -43,14 +44,27 @@ class SessionToken(BaseModel):
     jti: int
     sid: str
 
+    @root_validator()
+    def equal_iss_and_dest(cls, values: Dict[str, Any]):
+        domain = cls.__url_to_base(values.get('iss'))
+        try:
+            store_domain(domain)
+        except ValueError:
+            raise InvalidIssuerError(f"The domain {domain} is not a valid issuer.")
```
hillairet: Maybe a good opportunity for a `raise ... from`?

spylib/utils/session_token.py
```diff
@@ -62,58 +76,24 @@ def from_header(
...
-        """
-        Checks to be sure that the subject (sub) is the same as the user who made the request.
-        """
-        pass
+        # Verify enough fields specified and perform validation checks
+        return cls.parse_obj(payload)
 
-    def __url_to_base(self, url):
+    @classmethod
+    def __url_to_base(cls, url):
```
hillairet: You do not use or return `cls` so:
```suggestion
    @staticmethod
    def __url_to_base(url):
```

spylib/utils/session_token.py
```diff
@@ -43,14 +44,27 @@ class SessionToken(BaseModel):
...
+        domain = cls.__url_to_base(values.get('iss'))
+        try:
+            store_domain(domain)
+        except ValueError:
+            raise InvalidIssuerError(f"The domain {domain} is not a valid issuer.")
+        
+        if cls.__url_to_base(values.get('iss')) != cls.__url_to_base(values.get('dest')):
+            raise MismatchedHostError(
+                f'The issuer {cls.__url_to_base(values.get("iss"))} does not match '
+                f'the destination {cls.__url_to_base(values.get("dest"))}'
```
hillairet: Just a suggestion for readability for next time. No need to implement:
```suggestion
        iss_url = cls.__url_to_base(values.get('iss'))
        dest_url = cls.__url_to_base(values.get('dest'))
        if iss_url != dest_url:
            raise MismatchedHostError(
                f'The issuer {iss_url} does not match the destination {dest_url}'
```

hillairet (changes requested):
Minor cosmetic changes

spylib/utils/session_token.py
hillairet: I am confused ... This works perfectly with `mypy` for me. Maybe I'm not triggering the right problem
```python
class PropertyTest:
    def __init__(self):
        pass

    @property
    def prop(self) -> list:
        return [1, 2, 3, 4]

    @property
    def propstr(self) -> str:
        return 'Success'

protest = PropertyTest()

for i in protest.prop:
    print(i)

print(type(protest.propstr))
```
```bash
1
2
3
4
<class 'str'>
```

Nathan-Nesbitt: Sorry should have been more specific, for the type checking on the input of some of the functions I get errors:

`TypeError: startswith first arg must be str or a tuple of str, not property`

I can 'fix' this by casting the property to a string but that won't work as it doesn't actually contain the value. 

I found a fix, properties contain a method `fget()` which you can use to obtain the literal from the method.

Nathan-Nesbitt: Just kidding, better fix is to have to set it as a classmethod, which works for the default while a static method and method will not?

spylib/utils/session_token.py
```diff
@@ -0,0 +1,106 @@
...
+        secret: str,
+    ) -> SessionToken:
+
+        # Take the authorization headers and unload them
+        if not authorization_header.startswith(str(cls.prefix)):
+            raise TokenAuthenticationError(
+                "The authorization header does not contain a Bearer token."
+            )
+
+        token = authorization_header[len(str(cls.prefix)) :]
```
hillairet: What you need to [know about regex](https://www.python-course.eu/python3_re.php) and a [handy regex checker](https://regex101.com/)

```suggestion
        if result := re.search(r'Bearer (.*)', authorization_header):
            token = result.group(1)
        else:
            raise TokenAuthenticationError(
                "The authorization header does not contain a Bearer token."
            )
```

spylib/utils/session_token.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @classmethod
+    @property
+    def prefix(cls) -> str:
+        return "Bearer "
+
+    @classmethod
+    @property
+    def required_fields(cls) -> List[str]:
+        return ["iss", "dest", "sub", "jti", "sid"]
```
hillairet: As discussed I think this is overkill for this use case. Either have constants defined outside the class or just have the strings in the used once in the method. Not a big deal for something so unlikely to change.

hillairet (changes requested):
Remember that the convention is to use `'` and not `"`

spylib/utils/session_token.py
Nathan-Nesbitt: Is this just for performance? As if we use the raw string 'Bearer ' I feel like `startswith()` is faster/easier to read:

```python
if not authorization_header.startswith('Bearer '):
  raise TokenAuthenticationError(
    "The authorization header does not contain a Bearer token."
  )
```

spylib/utils/session_token.py
Nathan-Nesbitt: Changed it to constants as it will be easier to diagnose issues in the future

spylib/utils/session_token.py
hillairet: I like regexp but I might be the only one ... :sweat_smile: 
It's fine either way

spylib/utils/session_token.py
```diff
@@ -0,0 +1,95 @@
...
+class InvalidIssuerError(ValidationError):
+    pass
+
+
+class MismatchedHostError(ValidationError):
+    pass
+
+
+class TokenAuthenticationError(ValidationError):
+    pass
```
hillairet: As discussed let's avoid `ValidationError`.
Also to make them easier to find, it might be good practice to move these to a `session_token_exceptions.py` ... or not, I don't know.

```suggestion
class TokenValidationError(Exception):
    pass


class InvalidIssuerError(TokenValidationError):
    pass


class MismatchedHostError(TokenValidationError):
    pass


class TokenAuthenticationError(TokenValidationError):
    pass
```

hillairet (changes requested):

spylib/utils/session_token.py
Nathan-Nesbitt: If you don't mind, I wouldn't mind leaving them bundled with the object for now as it's the only thing throwing these errors. If we use them in the future in a different object, could we move them then?

hillairet (approved):
Well done! Let's chat about doing a release with that in. We need to test it in-situ as well. We may need a React dev for that!

Nathan-Nesbitt:
Thank you and sounds good @hillairet !

PR #39, opened 2021-10-17
Feature/add automated testing

hillairet:
Adding tox and github actions for automated testing. Currently all tests are passing. Nice!  :+1: 
The next step is to expand to older versions of python but the framework is here.


Nathan-Nesbitt (approved):
Looks good! Tox looks like a useful too, gonna add that to the mental library.

PR #40, opened 2021-10-19
Status code / request constants 

### LINKED ISSUE #27: Change REST goodstatus and error codes to class variables
Currently we accept a "goodstatus" parameter when executing rest:

https://github.com/SatelCreative/satel-spylib/blob/3bfc32f5eee8d632d85ebd5c54e37e3d9966d4a6/spylib/store.py#L120

Can we change this to be some token variable? Like

```
class Token:
  good_status: ClassVar[int] = 200
```

this would reduce the complexity of the method call and would reduce the chance someone makes a mistake.

### LINKED ISSUE #33: Replace **kwargs with `method` and any other required parameters
I am currently making requests to get orders based on the [Shopify REST docs](https://shopify.dev/api/admin-rest/2021-07/resources/order#[get]/admin/api/2021-07/orders/{order_id}.json), and have produced the following call to get an order:

```python
order = await token.execute_rest(goodstatus=200, debug=None, endpoint=f'/admin/api/2021-07/orders/{order_id}.json')
```

This is an error thrown by the `execute_rest` function if you call the method directly on an endpoint:

https://github.com/SatelCreative/satel-spylib/blob/b60c97ff2264473959ffd5b432b5fcddb535c327/spylib/token.py#L137-L160

At the moment the function expects `**kwargs` and runs an unchecked request to some endpoint based on those args:

https://github.com/SatelCreative/satel-spylib/blob/b60c97ff2264473959ffd5b432b5fcddb535c327/spylib/token.py#L143

I think from the README there is a required `method` parameter for this method?

Nathan-Nesbitt:
This was originally designed to solve issues in #27  but it incidentally solves issues associated with ambiguity in the input types for the rest requests (see #33)


spylib/utils/rest.py
```diff
@@ -0,0 +1,41 @@
+from enum import Enum
+
+from pydantic import BaseModel
+
+
+class Status(Enum):
```
hillairet: Hummmmm How about reusing [FastAPI's status](https://fastapi.tiangolo.com/tutorial/response-status-code/?h=status#shortcut-to-remember-the-names) which are actually just importing the [Starlette ones](https://github.com/encode/starlette/blob/master/starlette/status.py). Let's not reinvent the wheel. We already committed to fastapi in the OAuth Router anyway.

spylib/token.py
```diff
@@ -134,18 +135,30 @@ async def __handle_error(self, debug: str, endpoint: str, response: Response):
...
+            if not self.access_token:
+                raise ValueError("You have not initialized the token for this store. ")
+
+            response = await self.client.request(
+                method=request.method.value,
+                url=f'{self.api_url}{endpoint}',
+                headers={'X-Shopify-Access-Token': self.access_token},
+                json=json,
+            )
+            if response.status_code == Status.TOO_MANY_REQUESTS:
```
hillairet: I think you need to match the value, right?
```suggestion
            if response.status_code == Status.TOO_MANY_REQUESTS.value:
```

hillairet (commented):
Looking good.
While I like using Enums, maybe for this it's just best to reuse what already exists.

spylib/token.py
Nathan-Nesbitt: You're right, I'm surprised that didn't get caught by any of the tests.

spylib/utils/rest.py
Nathan-Nesbitt: Totally, will do that!

hillairet (changes requested):
Code looks good!
Docs missing.

Nathan-Nesbitt:
> Code looks good! Docs missing.

I just updated the docs on how to make a request. There may be a new issue required for reformatting the docs, as I was thinking it might be good to split up all of the `Token` methods that can be called into separate sections, and give a better explanation of how they work (along with examples).

hillairet (changes requested):
I didn't notice this before but there are not examples of querying Rest and GraphQL?
An example is always the best way to explain.

I'd say in both Rest and GraphQL:
1. Read one product
2. Change the title of one product

How does one pass the JSON payload in execute_rest???

Nathan-Nesbitt:
Ah I forgot I fixed the issue with **kwargs, I will add the docs for that! For the REST and GQL should we make it a separate issue?

README.md
```diff
@@ -89,33 +89,71 @@ token = OnlineToken(
...
-#### Querying Shopify
+### Querying Shopify
 
-We can query the store using either the REST endpoint or the GraphQL endpoint:
+#### REST
+
+We can query the store using either the REST:
+
+```python
+token.execute_rest(
```
hillairet:
```suggestion
async def token.execute_rest(
```

README.md
```diff
@@ -89,33 +89,71 @@ token = OnlineToken(
...
+  endpoint: str,
+  json: Optional[Dict[str, Any]],
+  debug: Optional[str],
+)
+```
+
+For example, if you want to query a product from shopify you can run:
+
+```python
+token.execute_rest(
```
hillairet:
```suggestion
product_json = await token.execute_rest(
```

README.md
```diff
@@ -89,33 +89,71 @@ token = OnlineToken(
...
+  request: GET,
+  endpoint: f'/products/{product_id}.json'
+)
+```
+
+If you want to update a product in a shop you can run:
 
 ```python
 token.execute_rest(
```
hillairet:
```suggestion
product_json = await token.execute_rest(
```

README.md
```diff
@@ -89,33 +89,71 @@ token = OnlineToken(
...
+Endpoint is the API endpoint string that we are querying, this should be similar to 
+the following format:
+
+```python
+f'/{resource}.json?fields={resource.param}&{[params]}'
+```
+
+The `debug` parameter is the message that is returned when there is an error. It is optional as it defaults to `""`.
 
+#### GQL
```
hillairet, line 144: This also deserves the same details as REST such as type of the parameters and an example.

hillairet (changes requested):
Docs tweaks.
Also removing the `**kwargs` means that we won't be able to pass things like timeout ... let's see through usage what is needed.

README.md
Nathan-Nesbitt: Do you mean `await`?

README.md
```diff
@@ -147,12 +147,58 @@ We can also query Shopify using the GraphQL endpoint:
...
 ```
 
+For example, if you want to query a product from shopify you can run:
+
+```python
+query = """
+{
+  shop {
+    name
+  }
```
hillairet: That will give you the Shop Name ... Could be good enough of an example though product query sounds better.

hillairet (changes requested):

hillairet (approved):

PR #45, opened 2021-10-26
Remove environment variables

### LINKED ISSUE #25: Question: Remove environment variable requirement
The OAuth requires that the user of the library knows that certain environment variables need to be set. Can we configure this so the user implements this portion in their application instead of forcing it in the library? 

https://github.com/SatelCreative/satel-spylib/blob/3bfc32f5eee8d632d85ebd5c54e37e3d9966d4a6/spylib/oauth/config.py#L4-L13

For example at the moment to check HMAC we pull in an environment variable:

https://github.com/SatelCreative/satel-spylib/blob/04cfd44ed8b3684512a76a5799a12f989bf91aae/spylib/oauth/validations.py#L40

Could we instead have some way to provide the arguments when we initialize the library?

Nathan-Nesbitt:
This removes the environment variable dependency from the library based on issue #25 . The logic for why this was done was:

1. *It is confusing to have environment variables in a library*:  You need to be super clear that they need to be set in the README, and you need to reflect these requirements in every new library / package that uses it.
2. *It limits the flexibility for the programmer*: I found that I would often need access to these variables, as they were defined in the library and not the app I would have to re-implement the code that was already done in spylib
3. *Conflicts*: If you have 2 libraries with the same variable name you are going to have problems as you won't be able to just change the env variable name as it is defined in the spylib library.

This is done by passing in the required environment variables to the oauth init function. 

https://github.com/SatelCreative/satel-spylib/blob/8c6686da95f6bcb9af4bbe8725dd66df8343ba97/spylib/oauth/router.py#L25-L37

Which later distributes the variables to other functions.

https://github.com/SatelCreative/satel-spylib/blob/8c6686da95f6bcb9af4bbe8725dd66df8343ba97/spylib/oauth/tokens.py#L18-L19


hillairet (changes requested):
Code looks great.
How about a little migration warning in the README for migrating to 0.3? Cause that's a breaking change, and not a small one.

README.md
```diff
@@ -127,7 +131,17 @@ You just need to call `init_oauth_router` such that:
...
+
+class ShopifyEnvironment(BaseSettings):
+    handle: str
+    api_key: str
+    secret_key: str
+
+    class Config:
+        env_prefix: str = 'shopify_'
 
+conf = ShopifyEnvironment()
```
hillairet: We don't need to tell people how to manage their environment variables.
```suggestion
```

README.md
```diff
@@ -144,6 +158,9 @@ oauth_router = init_oauth_router(
     user_scopes=['read_orders', 'write_products'],
     public_domain='my.app.com',
     private_key='KEY_FOR_OAUTH_JWT',
+    api_key=conf.api_key,
+    api_secret_key=conf.secret_key,
+    app_handle=conf.handle,
```
hillairet:
```suggestion
    api_key='SHOPIFY_APP_API_KEY',
    api_secret_key='SHOPIFY_APP_SECRET_KEY',
    app_handle='SHOPIFY_APP_HANDLE',
```

hillairet (changes requested):
Ok let's get this right in terms of docs.
Let's add how to get the App handle as described [here].(https://github.com/SatelCreative/st-pim/blob/9c31fbc6f6738e16444ea7d5cbb7fd0a2f063e32/server/config.sh.example#L9)
Also We should make the handle an optional parameter. But I can't find where to redirect to without it ... weird! Let's do that in another PR though.

Nathan-Nesbitt:
Created an issue for the optional handle @hillairet and fixed the rest!

hillairet (approved):
Cool let's push this through now.

PR #50, opened 2021-11-02
App Proxy Utils

### LINKED ISSUE #24: App proxy utils
Let's add the necessary tooling to handle [Shopify's app proxy calls](https://shopify.dev/apps/online-store/app-proxies).
As far as I know the only thing really needed is the HMAC validation of the signature.
What I don't know is if it is the same as the HMAC validation for oauth and webhooks? Are oauth and webhooks even using the same HMAC format????

Also it would be good to provide a function that could be used as [Dependency injection in the FastAPI endpoints](https://fastapi.tiangolo.com/tutorial/dependencies/#first-steps). Again I don't know if this should be specific to the service (app proxy, oauth, webhook) or if we can have a generic one for all services.

### LINKED ISSUE #41: Contain scope of functions for HMAC
This TLDR of the current process of validating the HMAC for a request during the oauth process:

1. The main check calls `validate_callback` 
2. `vaidate_callback` extracts the HMAC from the URL
3. `validate_callback` then calls `validate_callback_args` which strips the args
4. The args are then passed to `validate_hmac` (actually a rename of the `validate` function) 
5. The `validate_hmac` immediately calls `calculate_from_message` with the stripped information from the `validate_hmac`.

I think this could be simplified by having each function handle one core component of the process:

1. A function that just calculates the HMAC of a string (and check if it is encoded)
2. A function that takes a URL, strips the HMAC and passes it into function (1) returning the HMAC result
3. A final function that checks the requirements for a callback function, which passes a URL to (2)

Nathan-Nesbitt:
This provides a middle-ware solution to the app proxy HMAC check. I had to make some small changes to the HMAC validation functions, which happened to address another issue.


spylib/utils/hmac.py
```diff
@@ -12,16 +13,12 @@ def calculate_from_message(secret: str, message: str, is_base64: bool = False) -
...
-        path = path + '?' + query_string
-    message = path + datetime + body
-    return calculate_from_message(secret=secret, message=message, is_base64=is_base64)
+def validate_hmac(secret: str, message: Dict[str, list], is_base64: bool = False):
 
+    hmac_actual = message.pop('hmac')[0]
+    body = '&'.join([f'{arg}={",".join(message[arg])}' for arg in message.keys()])
 
-def validate(secret: str, sent_hmac: str, message: str, is_base64: bool = False):
+    message_hmac = calculate_message_hmac(secret, body, is_base64)
```
hillairet: Named arguments are preferred so it clearer and less error prone
```suggestion
    message_hmac = calculate_message_hmac(secret=secret, message=message, is_base64=is_base64)
```

hillairet (changes requested):
**These aren't small changes**. Whole code sections are removed.
What issue is being fixed??? The code looks broken to me.
The message is made out of the query arguments MINUS the `hmac` query arguments cause the hmac can't encode itself ... The management of this trick is now gone for no good reason.

spylib/oauth/validations.py
```diff
@@ -18,25 +13,8 @@ def validate_callback(shop: str, timestamp: int, query_string: Any, api_secret_k
...
-        validate_callback_args(args=original_args, api_secret_key=api_secret_key)
-
-
-def validate_callback_args(args: List[Tuple[str, str]], api_secret_key: str) -> None:
-    # We assume here that the arguments were validated prior to calling
-    # this function.
-    hmac_arg = [arg[1] for arg in args if arg[0] == 'hmac'][0]
-    message = '&'.join([f'{arg[0]}={arg[1]}' for arg in args if arg[0] != 'hmac'])
-    # Check HMAC
-    validate_hmac(secret=api_secret_key, sent_hmac=hmac_arg, message=message)
```
Nathan-Nesbitt, line 39: This functionality in this function was moved to `hmac.py` as the `validate_hmac` function wasn't doing anything before.

Nathan-Nesbitt:
@hillairet my newest change might explain why I made the changes to the HMAC function. Because the OAuth HMAC and the Proxy HMAC are so similar, I was trying to condense the existing code into 1 function instead of 2.

The HMAC value is popped out of the message in the `validate_hmac` stage.

tests/utils/test_app_proxy.py
```diff
@@ -0,0 +1,76 @@
...
+SHOP = 'test.myshopify.com'
+PREFIX = '/apps/api'
+TIMESTAMP = "1317327555"
+
+
+@pytest.fixture(scope='function', autouse=False)
+def url():
+    params = f'shop={SHOP}&path_prefix={PREFIX}&timestamp={TIMESTAMP}'
+
+    signature = calculate_message_hmac(API_SECRET, params)
```
hillairet, line 24: You shouldn't use our code to make the fixture that will test our code ... you are just checking consistency here, not accuracy!
Either copy paste values straight from Shopify or calculate them yourself "manually" with a different code once and hardcode here.
The first option is best but can be annoying to get.

spylib/utils/app_proxy.py
```diff
@@ -0,0 +1,41 @@
...
+
+class CheckAppProxy(BaseHTTPMiddleware):
+    def __init__(self, app: ASGIApp, secret: str, proxy_endpoint: str) -> None:
+        super().__init__(app)
+        self.secret: str = secret
+        self.proxy_endpoint: Path = Path(proxy_endpoint)
+
+    async def dispatch(self, request: Request, call_next):
+
+        # As middleware is run on everyt(hing, this allows us to avoid checking on
```
hillairet, line 22:
```suggestion
        # As middleware is run on everything, this allows us to avoid checking on
```

hillairet (changes requested):
This is the part of SPyLib that requires the best and most tests. So let's get it right.

The best might be to write some unit tests for the validation and HMAC calculation functions BEFORE changing the code. I should be in another PR merged onto main.
You could open another PR, make only tests on the old code, unit tests different from the one in this PR, then refactor.

Having a mistake in this part of the code means that the package is totally useless.

Nathan-Nesbitt:
Totally makes sense @hillairet I will write some tests and come back to this when they get merged!

Nathan-Nesbitt:
Created new tests for HMAC in #63, will continue once merged.

Nathan-Nesbitt:
@hillairet I'm trying to clean up the tests and the code so I can make this PR more simple, I should be able to re-use the HMAC functions but I am running into a snag as they are not as generic as expected.  

Do you know why we have a base64 check in the function? I can't seem to find any reference to it being used in the code, and if it isn't being used I think it might be best to remove this function:

https://github.com/SatelCreative/satel-spylib/blob/c9a02f148c6df662e229c8605db5f376b95a61ed/spylib/utils/hmac.py#L7-L14

If we need the base64 code, we can just check to see if it is base64, and handle it accordingly. This is because it is essentially just a function around another function without the base64 check.

**I will do this in a different issue and PR, then come back to this if you are okay with them.** I just think it will make it way clearer what is actually going on (and will allow me to reduce this PR to like 50 lines instead of 150)

hillairet:
@Nathan-Nesbitt So I found this base64 thing ... it's because the [webhook HMAC is base64 encoded](https://shopify.dev/apps/webhooks/https#verify-the-request-is-from-shopify)!

I have code that had a parameters `is_webhook` instead and I probably renamed to `is_base64` in case some other HMAC validation used a base 64. I can't think or find of one right now.
Should we pass an enum instead?

Nathan-Nesbitt:
> @Nathan-Nesbitt So I found this base64 thing ... it's because the [webhook HMAC is base64 encoded](https://shopify.dev/apps/webhooks/https#verify-the-request-is-from-shopify)!
> 
> I have code that had a parameters `is_webhook` instead and I probably renamed to `is_base64` in case some other HMAC validation used a base 64. I can't think or find of one right now. Should we pass an enum instead?

Okay should we have a parser for the webhook and the oauth which just handle this specific case? As I feel like this would be more clear if we just handle the base64 encoding in the specific instance (as it almost self documents that webhook HMAC are encoded) instead of having a shared function that tries to handle all the cases. Something like this?

**oauth/validations.py**
```python
def parse(message: Dict[str, list]):
    message.pop('hmac')[0]
    return '&'.join([f'{arg}={",".join(message[arg])}' for arg in message.keys()])

def validate(secret: str, hmac_actual: str, message: str):
    hmac_hash = hmac.new(secret.encode('utf-8'), message.encode('utf-8'), sha256).hexdigest()
    return hmac.compare_digest(hmac_actual, hmac_hash):
        
```

And we could have the following for the webhook handler in something like `utils/webhooks.py`:

```python
def validate(message: str):
    # Code for getting body of request
    hmac_header = request.headers.get('X-Shopify-Hmac-SHA256')
    digest = hmac.new(SECRET, data.encode('utf-8'), hashlib.sha256).digest()
    computed_hmac = base64.b64encode(digest)
    return hmac.compare_digest(computed_hmac, hmac_header.encode('utf-8'))
```

This would just remove a lot of functions and function calls (I think it would remove the entire `utils/hmac.py` file), and will make the code easier to follow.

hillairet:
I like it!
Indeed if it's only in webhooks as far as we know, let's keep the base64 in the webhook code.

lishanl:
Do we want to wrap this PR? @hillairet Do you want to pick it up from here or you have other suggestions?

hillairet:
There is some interesting clean up and new code here but we would do better with some planning and organizing from scratch I think.
I'd like some more input on the best way to organize the hmac validation actually.

PR #55, opened 2021-11-03
Oauth Get Token Update

### LINKED ISSUE #44: Bundle function into method
Ran into this while I was cleaning up another issue:

https://github.com/SatelCreative/satel-spylib/blob/04cfd44ed8b3684512a76a5799a12f989bf91aae/spylib/oauth/tokens.py#L16-L19

This function is called in only one place; this class:

https://github.com/SatelCreative/satel-spylib/blob/04cfd44ed8b3684512a76a5799a12f989bf91aae/spylib/oauth/tokens.py#L41-L53

This should probably just be a method, as it will reduce the possibility of someone importing it, and it will be easier to read (as a private class method will indicate that only this class uses this method, and it should not be called publicly).

Nathan-Nesbitt:
This removes the duplicate implementation of the `OfflineToken` and `OnlineToken` classes.


spylib/token.py
```diff
@@ -26,6 +26,18 @@
     ShopifyThrottledError,
     not_our_fault,
 )
+from spylib.utils.domain import domain_to_storename
+
+from .utils import HTTPClient
+
+
+class TokenResponse(BaseModel):
```
Nathan-Nesbitt, line 34: I added this in so I could use it as a parent type to the `get_token` function.

spylib/token.py
```diff
@@ -282,3 +300,45 @@ def load(cls, store_name: str, associated_user: str):
...
+        message = (
+            f'Problem retrieving access token. Status code: {response.status_code} {jsondata}'
+            f'response=> {response.text}'
+        )
+        logger.error(message)
+        raise ValueError(message)
+
+    jresp = Response.parse_obj(response.json())
+
+    # We have to use this syntax
```
Nathan-Nesbitt, line 340: This merges the dict and the store name value.

tests/oauth/test_base.py
```diff
@@ -3,50 +3,19 @@
...
-    associated_user={
-        'id': 902541635,
-        'first_name': 'John',
-        'last_name': 'Smith',
-        'email': 'john@example.com',
-        'email_verified': True,
-        'account_owner': True,
-        'locale': 'en',
-        'collaborator': False,
-    },
```
Nathan-Nesbitt, line 49: Removed because they are duplicate definitions from `token_classes.py`.

tests/token_classes.py
```diff
@@ -15,16 +16,16 @@
...
-    associated_user={
-        'id': 902541635,
-        'first_name': 'John',
-        'last_name': 'Smith',
-        'email': 'john@example.com',
-        'email_verified': True,
-        'account_owner': True,
-        'locale': 'en',
-        'collaborator': False,
-    },
```
Nathan-Nesbitt, line 27: There was already an AssociatedUser type defined in the OAuth section, so I reverted the object back to using it instead of a dict

tests/oauth/test_base.py
```diff
@@ -65,106 +34,124 @@ async def test_oauth(mocker):
...
+        test_information.app_scopes,
+        test_information.user_scopes,
+        test_information.public_domain,
+        test_information.private_key,
+        test_information.app_handle,
+        test_information.api_key,
+        test_information.api_secret_key,
+        OfflineToken,
+        OnlineToken,
+    )
```
Nathan-Nesbitt, line 48: Python treats class variables that point to other classes as subclasses, even if they are variables. This means I cannot set OfflineToken and OnlineToken in the main `test_information` class, which in turn means I cannot just pass in the object as kwargs.

spylib/token.py
```diff
@@ -282,3 +300,45 @@ def load(cls, store_name: str, associated_user: str):
...
         therefore the developer should override this.
         """
+
+
+async def get_token(
+    domain: str,
+    code: str,
+    api_key: str,
+    api_secret_key: str,
+    CustomToken: Union[Type[OfflineTokenABC], Type[OnlineTokenABC]],
```
Nathan-Nesbitt, line 310: Because we don't define `save` and `load` we cannot simply treat the input or return values as type `Token`. Technically we could just define abstract methods in `Token` but this would break the `SOLID` principles you were talking about

spylib/token.py
```diff
@@ -282,3 +300,45 @@ def load(cls, store_name: str, associated_user: str):
...
+        logger.error(message)
+        raise ValueError(message)
+
+    jresp = Response.parse_obj(response.json())
+
+    # We have to use this syntax
+    return CustomToken(
+        **{'store_name': domain_to_storename(domain)},
+        **jresp.dict(),
+    )
```
Nathan-Nesbitt, line 344: This merges the dict and the store name value. Because the response from the API misses the store name.

spylib/oauth/__init__.py
```diff
@@ -1,2 +1,3 @@
 from .router import init_oauth_router  # noqa: F401
```
hillairet, line 1:
```suggestion
from .router import init_oauth_router
```

spylib/token.py
```diff
@@ -265,10 +270,23 @@ class OnlineTokenABC(Token, ABC):
...
+    associated_user: AssociatedUser
+    associated_user_scope: List[str]
     expires_in: int = 0
     expires_at: datetime = datetime.now() + timedelta(days=0, seconds=expires_in)
 
+    @validator('associated_user', pre=True, always=True)
+    def associated_user_convert(cls, v):
+        if type(v) == dict:
+            return AssociatedUser(**v)
+        return v
```
hillairet, line 282: Why isn't pydantic doing this by itself???

spylib/oauth/router.py
```diff
@@ -115,29 +122,28 @@ async def shopify_callback(request: Request, args: Callback = Depends(Callback))
...
-        )
+        if OnlineToken:
+            online_token = await get_token(
+                domain=args.shop,
+                code=args.code,
+                api_key=api_key,
+                api_secret_key=api_secret_key,
+                CustomToken=OnlineToken,
+                Response=OnlineTokenResponse,
+            )
```
hillairet, line 133: Following the same approach as all our current other Shopify API calls, let's keep this function simple and have it just return JSON. It's role is to abstract the API call, not check the JSON itself, just getting it.
The Response class thing is a neat idea but it does the exact same validation as the OnlineToken, one line prior. That's overly complicated.
It seems like `get_token` doesn't even change if it's online or offline token ...

Now what you can do is catch the pydantic error if the JSON doesn't match what we want and add a message to clarify that!

 ```suggestion
             online_token_json = await get_token(
                 domain=args.shop,
                 code=args.code,
                 api_key=api_key,
                 api_secret_key=api_secret_key
             )
             online_token_json['store_name'] = domain_to_storename(args.shop)
             online_token = OnlineToken.parse_obj(online_token_json)
```

hillairet (changes requested):
Sorry I'm :hankey:ing in your idea to make fancy code again. I just want to keep it simple and consistent with the code we have in our existing apps. Your solution has a lot of redundancy or I totally misunderstood it.

lishanl:
Do we still want to move forward with this PR? @hillairet

hillairet:
Since this is not starting from a clear issue and I'm not sure where it was going, let's just start over.

PR #63, opened 2021-11-25
Added HMAC tests

### LINKED ISSUE #62: Create tests for HMAC functions
Based on #54 and #50 there needs to be some tests to ensure the validity and consistency of the HMAC functions currently in the repo.

Nathan-Nesbitt:
Solves #62


hillairet (approved):
Some of the most useful tests we can have!
What happened to the bug you struggled with the other day?

Nathan-Nesbitt:
@hillairet oh the JWT issue with validation? I realized that I hadn't properly generate the public/private keys so when I deployed it was failing on all of the tokens!

PR #65, opened 2021-11-25
Added gh action to compare files

### LINKED ISSUE #57: Create a github action that would compare docs/index.md and README.md
Make sure that both of the files are identical and the action should fail if not.

jsemearns:
- Added new action to compare files (README.md and docs/index.md)

Fix #57

### DIFF
--- a/.github/workflows/compare.yml
+++ b/.github/workflows/compare.yml
@@ -1,0 +1,28 @@
+name: Compare index and readme
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.9"]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test files
+        run: |
+          DIFF=$(diff README.md docs/index.md)
+          if [ "$DIFF" != "" ]; then
+              exit 1
+          else
+              exit 0
+          fi
--- a/docs/index.md
+++ b/docs/index.md
@@ -90,33 +90,116 @@ token = OnlineToken(
 )
 ```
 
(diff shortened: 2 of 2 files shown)


docs/index.md
```diff
@@ -222,12 +305,15 @@ because unlike poetry it allows to set the metadata on pypi such as author or ho
...
-2. Commit to git
-3. Run `poetry build` to create the package folders in `/dist`
-4. Run `flit publish` to publish to PyPI
-5. Tag the release in git and push it to Github
+3. Commit to git
+4. Open PR for other codeowners to review
+5. Once approved, squash and merge `release/x.y.z`
+6. Run `poetry build` to create the package folders in `/dist`
+7. Run `flit publish` to publish to PyPI
+8. Tag the release in git and push it to Github
```
Nathan-Nesbitt, line 316: Thanks for updating these instructions! Super helpful!

.github/workflows/compare.yml
```diff
@@ -0,0 +1,28 @@
+name: Compare index and readme
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
Nathan-Nesbitt, line 9: Not crazy well versed in GitHub actions, but will this not update the docs when there is a PR that hasn't been merged? If so, is this what we want because the docs may not have been approved?

jsemearns: This action is just so that we make sure both README and docs/index.md are identical on every PR to main. We don't really automate the update of either of the two files as there can be problems with it.

Nathan-Nesbitt: Oh okay sweet @jsemearns looks great!

Nathan-Nesbitt (approved):

hillairet (approved):
I think I forgot to press Approve twice on this PR after looking at it!
Sorry for the delay  :sweat_smile:

PR #69, opened 2021-12-03
Support/0.3

### LINKED ISSUE #68: Pip not installing spylib dependencies
Pip does not appear to install the dependencies of spylib. Steps to reproduce:

```
mkdir /tmp/spylib-test
cd /tmp/spylib-test

virtualenv env
source env/bin/activate.fish

pip install spylib
python

import spylib
```

Received error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/spylib-test/env/lib/python3.9/site-packages/spylib/__init__.py", line 6, in <module>
    from .token import OfflineTokenABC, OnlineTokenABC, Token
  File "/tmp/spylib-test/env/lib/python3.9/site-packages/spylib/token.py", line 8, in <module>
    from httpx import AsyncClient, Response
ModuleNotFoundError: No module named 'httpx'
```

I believe this may be due to one of two things:
- Not exporting a `requirements.txt` file before publishing
- The generated `setup.py` not containing the `install_requires` kwarg

hillairet:
Alright so proper publication process to pypi!
Let's use Poetry from start to finish because it can do it all, at least what we need.
I thought we need flit for the pypi metada but we don't so all gooood!

We still need to update the README in main to reflect that we don't use flit anymore.

Fix #68

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "spylib"
-version = "0.3"
+version = "0.3.1"
 description = "A library to facilitate interfacing with Shopify's API"
 authors = ["Anthony Hillairet <ant@satel.ca>"]
 license = "MIT"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "spylib"
-version = "0.3.1"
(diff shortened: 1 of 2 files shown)


hillairet:
I have to merge so I can release 0.4

PR #73, opened 2021-12-07
Release/0.4

hillairet:
Creating the release 0.4 because it has useful changes we need.
Main request we got is to remove the environment variable need for OAuth.


hillairet:
@Nathan-Nesbitt 
I'll merge now to be able to make a quick release to unblock Quinn

PR #76, opened 2022-01-05
Fix session token format

### LINKED ISSUE #75: Session tokens issues
First, the new form of the session tokens does not appear to match the docs `README.md`. I attempted to use depends however there seems to be no mechanism in the function itself to extract a header from the request? The method was also called `from_header` rather than `decode_token_from_header` which was referenced in the docs

A suggestion is to write tests based on your examples in the README - that way if they break it will be clear what needs to be updated

An issue I am having to work around now is that shopify does not actually pass the `jti` as an int. At least in the session token I received it was a string ('3512a085-ee9a-4914-b252-3aabcd1ada14')

For your convenience here is a full session token payload:

```js
{
  "iss": "https://satel-products.myshopify.com/admin",
  "dest": "https://satel-products.myshopify.com",
  "aud": "2ca57b9f8ea64fc449ca4d20610884ec",
  "sub": "78980514043",
  "exp": 1638925238,
  "nbf": 1638925178,
  "iat": 1638925178,
  "jti": "3512a085-ee9a-4914-b252-3aabcd1ada14",
  "sid": "38a01da7a45fbc47b07ddee21155e017d3e3e04b646e766835135021d24e39ee"
}
```

Edit:

This workaround seems to work:

```python
class MySessionToken(SessionToken):
    jti: str
```

Nathan-Nesbitt:
The session token format was not described properly in the README. This has been updated to illustrate the whole process, including a new test which should match the process of pulling the token from the header.

### DIFF
--- a/tests/utils/test_session_token.py
+++ b/tests/utils/test_session_token.py
@@ -1,7 +1,11 @@
 from datetime import datetime, timedelta
+from functools import partial
 
 import jwt
 import pytest
+from fastapi import Depends, FastAPI
+from fastapi.testclient import TestClient
+from starlette.requests import Request
 
 from spylib.utils.session_token import (
     InvalidIssuerError,
@@ -28,7 +32,7 @@ def get_token():
         'exp': (now + timedelta(0, 60)).timestamp(),
         'nbf': (now - timedelta(0, 60)).timestamp(),
         'iat': now.timestamp(),
-        'jti': 123,
+        'jti': '3512a085-ee9a-4914-b252-3aabcd1ada14',
         'sid': 'abc123',
     }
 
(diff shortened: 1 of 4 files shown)


tests/utils/test_session_token.py
```diff
@@ -80,3 +83,22 @@ async def test_invalid_token_parameters(token, parameter, value, error):
...
+
+    @app.get('/token')
+    async def token_endpoint(token: SessionToken = Depends(parse_session_token)):
+        return token
+
+    client = TestClient(app=app)
+
+    header = generate_auth_header(token)
+
+    client.get('/token', headers={'Authorization': header})
```
hillairet, line 104: What is the actual test? Is it just checking that the code is running without exception?
I guess that's enough

hillairet (approved):

tests/utils/test_session_token.py
Nathan-Nesbitt: I agree, it was more of a sanity check to show that the dependency works (they're quite touchy).

Nathan-Nesbitt:
@jsemearns how do we get the `compare index and readme` to pass?

hillairet:
@Nathan-Nesbitt The main README.md and [docs/index.md](https://github.com/SatelCreative/satel-spylib/blob/fix/session-token-format/docs/index.md) should be identical.
You modified the README.md so now you need the exact same changes in docs/index.md

Nathan-Nesbitt:
> @Nathan-Nesbitt The main README.md and [docs/index.md](https://github.com/SatelCreative/satel-spylib/blob/fix/session-token-format/docs/index.md) should be identical. You modified the README.md so now you need the exact same changes in docs/index.md

Oh didn't know! Easy enough. Could we write a GitHub actions that overwrites the `docs/index.md` with `README.md` on push? Cause then we wouldn't have to worry about checking, it would always be up to date @hillairet .

Nathan-Nesbitt:
I am going to keep the original issue open as I'm waiting for a reponse, but I think this solves a portion of the issues and can be merged.

qw-in:
I'll give this a try, about to work on the utfs server again. Thanks!

PR #79, opened 2022-01-24
Add 3.8 and 3.10 to tests

### LINKED ISSUE #70: Add versions 3.8 and 3.10 to the standard testing
Since we have the whole system to test different versions of python, we should test at least for 3.10.
It would be nice to have 3.8 but only if it doesn't require too much work on the requirements and code.

Nathan-Nesbitt:
fix #70

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,10 @@
 [tox]
 isolated_build = true
-envlist = py39, py310, flake8, mypy, black
+envlist = py38, py39, py310, flake8, mypy, black
 
 [gh-actions]
 python =
+    3.8: py38, flake8, mypy, black
     3.9: py39, flake8, mypy, black
     3.10: py310, flake8, mypy, black
 
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,9 @@
 [tox]
 isolated_build = true
-envlist = py38, py39, py310, flake8, mypy, black
+envlist = py39, py310, flake8, mypy, black
 
 [gh-actions]
 python =
-    3.8: py38, flake8, mypy, black
     3.9: py39, flake8, mypy, black
     3.10: py310, flake8, mypy, black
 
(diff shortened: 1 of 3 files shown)


Nathan-Nesbitt:
@rahulpatidar0191 any idea how to fix the missing windows binary issues displayed in the tests?

hillairet:
Weird error.
Don't worry too much about Windows.
We can remove the windows test for now if needed and add them back later. Better to have more python versions tested.

Nathan-Nesbitt:
@hillairet I think this is an error with mypy and just Windows with python 3.10 as 3.9 still works. I'll remove windows for now

Nathan-Nesbitt:
3.8 might be an issue, as many of our dependencies are not compatible with 3.8 should we keep them?

hillairet (changes requested):
Ok sweet: we learned that our current set of dependencies are not compatible with python 3.8! And that's good to know.
Fine for now. We can make an issue to try to recover the compatibility if possible.

Could you make a note in the README somewhere towards the top?
Also it'd be nice to have a badge in the README to show which versions work. We could potentially have a red badge for 3.8 ...
If it's easy to do the badge, do it, otherwise we'll do it later.

tox.ini
```diff
@@ -1,10 +1,12 @@
 [tox]
 isolated_build = true
-envlist = py39, flake8, mypy, black
+envlist = py38, py39, py310, flake8, mypy, black
 
 [gh-actions]
 python =
+    3.8: py38, flake8, mypy, black
     3.9: py39, flake8, mypy, black
+    3.10: py310, flake8, mypy, black
```
hillairet: The static code analysis shouldn't be python version dependent. Let's just run it for the latest python.
We only need to run pytest for each python version:
```suggestion
    3.8: py38
    3.9: py39
    3.10: py310, flake8, mypy, black
```

hillairet (changes requested):

hillairet (approved):

tox.ini
lishanl: right, I remembered you said something about this but didn't remember what exactly, so this is it. 🆒

PR #81, opened 2022-02-25
Update Shopify API

ponty33:
Update Shopify API version to `2022-01`

### DIFF
--- a/docs/api_version.md
+++ b/docs/api_version.md
@@ -1,0 +1,27 @@
+# Change Shopify API version
+- As shown in the `Implement Token Classes`, token can be used as following:
+```python
+from spylib import OfflineTokenABC
+
+
+class OfflineToken(OfflineTokenABC):
+  async def save(self):
+      pass
+
+  @classmethod
+  async def load(cls, store_name: str):
+      pass
+```
+
+- If the Shopify API version used in Spylib is not the desired version, it can be changed with the following:
+```python
+class OfflineToken(OfflineTokenABC):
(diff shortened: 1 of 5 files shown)


docs/api_version.md
```diff
@@ -0,0 +1,27 @@
...
+```python
+class OfflineToken(OfflineTokenABC):
+  # Add the version to be used here
+  api_version: '2022-01'
+  async def save(self):
+      pass
+
+  @classmethod
+  async def load(cls, store_name: str):
+      pass
```
hillairet: So I tried that way of changing the class variable using a simple example and I got this error `Field name "cv" shadows a BaseModel attribute; use a different field name with "alias='cv'".`.

I think [this TWIL](https://github.com/orgs/SatelCreative/teams/python/discussions/152) is relevant

```suggestion
class OfflineToken(OfflineTokenABC):
  async def save(self):
      pass

  @classmethod
  async def load(cls, store_name: str):
      pass
      
# Add the version to be used here
OfflineToken.api_version: '2022-01'
```

hillairet (changes requested):
Fix docs

docs/api_version.md
ponty33:
```python
OfflineToken.api_version: '2022-01'
```

The code is working, but I wonder why I got this flake8 error from the line above
`syntax error in forward annotation '2022-01'flake8(F722)`

hillairet (approved):

PR #82, opened 2022-03-03
Add path_prefix param to init oauth router and url

### LINKED ISSUE #74: Cannot mount oauth router with a prefix
Because of the way `callback_path` is used for both mounting a route within the oauth_router and as the path appended to the redirect url any prefix means these get out of sync

```python
oauth_router = init_oauth_router(
    callback_path='/oauth/callback',
)

app.include_router(
    oauth_router,
    prefix="/api"
)
```

In this configuration Shopify attempts to redirect back to `my.domain/oauth/callback` which is not a valid route

For now my workaround is to not use a prefix for the router, but this is not a great long term solution

lishanl:
Enable the setting of prefix for the oauth routes. 
fix: https://github.com/SatelCreative/satel-spylib/issues/74


hillairet (changes requested):
Do you mind adding this to the documentation too? Important so we advertise it.

hillairet (approved):

PR #85, opened 2022-03-07
Fix/dependency versions

### LINKED ISSUE #77: Dependency versions are too strict
Especially the packages that have not hit `1.0.0` yet. At the moment spylib insists on specifically "fastapi@0.68". [Here is some helpful documentation from poetry that could help make it a little more flexible](https://python-poetry.org/docs/dependency-specification/)

hillairet:
Fix #77 

After properly reading the poetry docs I now realize that the default caret is not right for our purpose.
I also took this opportunity to remove the lock file because poetry's own documentation says to not put it in the VCS for libraries.


qw-in (approved):
Awesome, thanks!

lishanl (approved):

PR #88, opened 2022-03-08
📝 Reorganize the documentation in sections

hillairet:
Use sections per Shopify component handled in SPyLib to make its features clearer.

Some more work is needed. This is a first pass.


docs/admin-api.md
```diff
@@ -0,0 +1,204 @@
...
+  async def save(self):
+      # Some code to save the token to a database
+
+  @classmethod
+  async def load(cls, store_name: str, user_id: str):
+      # Some code to load the token from the database
+
+class PrivateToken(PrivateTokenABC):
+  @classmethod
+  async def load(cls, store_name: str, user_id: str):
```
lishanl, line 32: just saw this, probably don't need `user_id` for private app token?

docs/development-contributing.md
```diff
@@ -0,0 +1,64 @@
...
+--output-path="./docs/api-docs" \
+--validate spylib
+
+mkdocs build
+mkdocs serve
+```
+
+The default URL is at `127.0.0.1:8000`.
+
+## Publishing to pipy
```
lishanl, line 48: seems a typo, you mean pypi?

lishanl (approved):
Thanks Ant.
Just a thought, maybe we want to be consistent when referring to SPyLib throughout the doc.  we have used SPyLib and Spylib. Also, for things like Shopify and PyPI.

docs/admin-api.md
hillairet: Please make an issue to review this properly since it's not the point of this PR.

lishanl: though it does look like just a documentation error since the code doesn't have the `user_id` param
https://github.com/SatelCreative/spylib/blob/main/spylib/token.py#L311

lishanl: We could have separate PR to fix documentation error if any since this PR is strictly reorganizing the content?

PR #95, opened 2022-03-11
Small config changes

qw-in:
- Pin development dependencies
- Add brunette and remove flake8-quotes
- Add .vscode settings to automatically configure editor
- Merge settings into pyproject.toml/setup.cfg

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,26 @@
+{
+  // hide cache files
+  "files.exclude": {
+    "**/__pycache__": true,
+    "**/.mypy_cache": true,
+    "**/.pytest_cache": true,
+  },
+  // brunette & black
+  "editor.formatOnSave": true,
+  "python.formatting.provider": "black",
+  "python.formatting.blackPath": "brunette",
+  // isort
+  "[python]": {
+    "editor.codeActionsOnSave": {
+      "source.organizeImports": true
+    }
+  },
+  // mypy
+  "python.linting.mypyEnabled": true,
+  // pytest
+  "python.testing.unittestEnabled": false,
+  "python.testing.pytestEnabled": true,
+  "python.testing.pytestArgs": [
+    "tests"
+  ],
+}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,8 @@ license = "MIT"
 readme = "README.md"
 exclude = ["tests", "scripts"]
 classifiers = [ "License :: OSI Approved :: MIT License",]
-homepage = "https://github.com/SatelCreative/satel-spylib"
+homepage = "https://satelcreative.github.io/spylib"
+repository = "https://github.com/SatelCreative/spylib"
 
 [tool.poetry.dependencies]
 python = "^3.8"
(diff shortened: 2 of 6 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,26 @@
+{
+  // hide cache files
```
lishanl, line 2: The settings look pretty clear to me so maybe the comments are not necessary.

.vscode/settings.json
```diff
@@ -0,0 +1,26 @@
...
+  "files.exclude": {
+    "**/__pycache__": true,
+    "**/.mypy_cache": true,
+    "**/.pytest_cache": true,
+  },
+  // brunette & black
+  "editor.formatOnSave": true,
+  "python.formatting.provider": "black",
+  "python.formatting.blackPath": "brunette",
+  // isort
```
lishanl, line 12: just referring to the doc here: https://github.com/SatelCreative/skr/blob/main/python/python_vscode.md
maybe adding following makes it more specifically use `isort`?
```json
    "python.sortImports.args": [
        "-sp .isort.cfg"
    ],
```

pyproject.toml
```diff
@@ -54,3 +55,12 @@ skip-string-normalization = true
...
 known_third_party = ["fastapi", "pydantic", "starlette"]
+
+[tool.mypy]
+plugins = "pydantic.mypy"
+show_error_codes = true
+
+# https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file
+[[tool.mypy.overrides]]
+ignore_missing_imports = true
+module = []
```
lishanl, line 66: I guess this would be 
```
module = ["spylib.tests.*"]
```
based on the original mypy setting?

lishanl (commented):
Good call to pin the dev dependencies

hillairet (approved):

.vscode/settings.json
qw-in: I can remove them. It does always feel weird using json5 in a `.json` file anyway

.vscode/settings.json
qw-in: We actually don't have an isort.cfg. If seems to be detecting the config for me - maybe on windows it isn't working?

pyproject.toml
qw-in: Mypy seemed fine without ignoring the errors so I left it out. @hillairet do you know why the test module was ignored?

.vscode/settings.json
lishanl: Right, we don't have an `isort.cfg` file. 😅  I use WSL for `spylib`, it automatically sorts the import for me on save. 
I guess for the future, whether to standardize the isort settings as well. e.g. used in the [satel-webapp-boilerplate](https://github.com/SatelCreative/satel-webapp-boilerplate/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D/server/.isort.cfg) [multi_line_output = 3](https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html#3-vertical-hanging-indent)

.vscode/settings.json
lishanl: oh.. [JSON5](https://json5.org/), new to me. 🤓 
maybe only the 
```
// isort
```
comment is necessary since it's not mentioned in the settings.

pyproject.toml
lishanl: I did encounter some typing error about `Box` when I ran the tests the first time. I modified the module there and the error went away. I reverted the change and ran the tests again, the previous typing error didn't come back. So I don't know either.. haha 😕

qw-in: You may need to tell mypy not to use the cache - I've had issues with that before

hillairet: I don't remember the reason to avoid running mypy on the tests.
It'd be better to run mypy on them obviously though.
Right now mypy is happy in the Github action.
@lishanl Maybe you need a `poetry update`? Maybe the proper typing for added to `Box` or something?

qw-in:
Going to merge! I prefer to have the comments than not - hope that is ok @lishanl

lishanl:
> Going to merge! I prefer to have the comments than not - hope that is ok @lishanl

yea. That helps make the settings obvious without the need to look at the actual settings. It would help when the settings get longer. It's cool. Thanks Quinn.

PR #98, opened 2022-03-11
:bug: Fix tests to prevent pytest from hanging

### LINKED ISSUE #94: Remove fastapi version upper limit
For some reason the github action for python 3.9 and 3.10 would just hang if a fastapi versoion 0.69 and above was used. See #85 for the tests.

This must be related somehow to the introduction of Trio in fastapi 0.69 somehow.

hillairet:
Fix #94 

It looks like it was some old "standard" code that I can't remember why we needed it before but I suspect that the pytest async library is now doing what the conftest file was doing before.

Locally the problem is fixed for python 3.10


qw-in (approved):
Assuming you checked with 0.68 and below right? Nice job!

hillairet:
@jsemearns I keep thinking this is Slackers so I put you as reviewer at first. Feel free to review though!  :wink:

hillairet:
> Assuming you checked with 0.68 and below right? Nice job!

Oh no I didn't think of that. It does not work ... :persevere: 
But I don't see a problem with requiring `>= 0.69.0`.
Do you foresee any issue? The version 0.69.0 is 5 months old now ...

lishanl (approved):
Hard to see the interrelations, another good reason to identify and remove the unnecessary dependency

pyproject.toml
```diff
@@ -12,12 +12,11 @@ homepage = "https://github.com/SatelCreative/satel-spylib"
 
 [tool.poetry.dependencies]
 python = "^3.8"
-fastapi = ">= 0.65.1, < 0.69.0"
+fastapi = ">= 0.65.1"
```
qw-in:
```suggestion
fastapi = ">= 0.69.0"
```

qw-in (changes requested):
Yeah, I don't see any issues. As long as we release it as `0.5` or something and note the breaking change

pyproject.toml
hillairet: Definitely 0.5

qw-in (approved):

PR #101, opened 2022-03-15
Make fastapi optional and raise error when imported but not installed

### LINKED ISSUE #84: Reduce dependencies
Our current dependencies were chosen out of habit and convenience without considering that this is a library so we should have a minimal set of dependencies:

- [x] Raise an exception with a clear message to install fastapi whenever `spylib/oauth/router.py` is loaded and fastapi cannot be imported.
- [x] Remove loguru and use the standard library logger instead.
- [x] Copy paste code from the shortuuid library that we need to make our `get_unique_id` independent of shortuuid
- [x] Move nest-asyncio to dev dependencies (must have been put as main dependency by accident)

lishanl:
fix: #84  - Raise an exception with a clear message to install fastapi whenever spylib/oauth/router.py is loaded and fastapi cannot be imported.

First Pass: 
- Though `fastapi` is not required, `pydantic` and `starlette` are. They were installed when installing `fastapi`, now added to the required dependencies instead. 
- Raise import error when `spylib.oauth.router` is accessed without `fastapi` installed
- Run all tests like before if `fastapi` is installed, else skip the tests that use `fastapi`


lishanl:
Hey @hillairet , can you help please? I can't get tox to install also the optional dependencies `fastapi` to run all the tests. Do you know how? Thanks

hillairet:
Ok I tried the fancy way and it wasn't installing the dependency so I made it to another test environment entirely and it's working great!

First the `py310` test env with the fastapi tests skipped as currently intended:
image

Then the `fastapi` test env with fastapi installed, all the tests run:
image


Also to tox just do:
```
poetry shell
tox
```
It will complain about the missing python version but that's OK.
Well if you have python 3.9, you will run only pytest without fastapi.

hillairet:
@lishanl Another thing!
Instead of skipping the tests, it would be much better to check that the exception is thrown.

So within the test, check if fastapi is available:
* If it is, run the current test
* If it isn't, load the router and use the assert on exception to catch and verify the exception.

lishanl:
> @lishanl Another thing! Instead of skipping the tests, it would be much better to check that the exception is thrown.
> 
> So within the test, check if fastapi is available:
> 
> * If it is, run the current test
> * If it isn't, load the router and use the assert on exception to catch and verify the exception.

Good suggestion. thanks @hillairet

tests/oauth/test_base.py
```diff
@@ -69,97 +67,104 @@ def json(self):
...
-            'code': 'LOGINCODE',
-        },
-    )
-
-    TEST_DATA.post_login.assert_called_once()
-    TEST_DATA.post_login.assert_called_with('test', OnlineToken(**ONLINETOKEN_DATA))
+    with pytest.raises(FastAPIImportError):
+        from spylib.oauth import OfflineToken, OnlineToken, init_oauth_router
+
+        app = FastAPI()
```
hillairet: I think there is a more readable way to do this that avoids having `except ImportError: pass` which is a bit of a no-no, at least I don't want to give a precedent for it.

Also avoids the big change in indentation ... always nice for git diff :wink: 

```suggestion
    if 'fastapi' not in sys.modules and importlib.util.find_spec('fastapi') is None:
        with pytest.raises(FastAPIImportError):
            from spylib.oauth import OfflineToken, OnlineToken, init_oauth_router
        return

    from spylib.oauth import OfflineToken, OnlineToken, init_oauth_router
    
    app = FastAPI()
```

hillairet (changes requested):
So on the right track.
I put a code suggestion for the case of having one test but I think I know the right way to do this!

Let's have two independent tests and skip the one(s) that is not applicable for each situation.

1. Test when fastapi is not installed; skip test when fastapi is installed
2. Test when fastapi is installed; skip test when fast is not installed

This way the list of tests will be explicit:
```
[test_oauth_router_without_fastapi]: passed
[test_oauth_router_with_fastapi]: skipped
```
or
```
[test_oauth_router_without_fastapi]: skipped
[test_oauth_router_with_fastapi]: passed
```

Python zen: "Explicit is better than implicit."   :wink:

scripts/test_suite.sh
```diff
@@ -12,7 +12,7 @@ reportvalidation() {
 }
 
 
-python -m pytest -vv --durations=3 --cov=spylib --cov-report term-missing
+python -m pytest -rs -vv --durations=3 --cov=spylib --cov-report term-missing
```
hillairet: Why `-rs`?

spylib/oauth/router.py
```diff
@@ -2,7 +2,12 @@
 from inspect import isawaitable
 from typing import Awaitable, Callable, List, Optional, Union
 
-from fastapi import APIRouter, Depends, HTTPException, Query
+from spylib.exceptions import FastAPIImportError
+
+try:
+    from fastapi import APIRouter, Depends, HTTPException, Query  # type: ignore
+except ImportError as e:
+    raise FastAPIImportError('fastapi is not installed, run `pip install spylib[fastapi]`') from e
```
hillairet:
```suggestion
    raise FastAPIImportError('The oauth router is a fastapi router and fastapi is not installed. Run `pip install spylib[fastapi]` to be able to use the oauth router.') from e
```

tests/oauth/test_base.py
```diff
@@ -61,7 +61,20 @@ def json(self):
 
 
 @pytest.mark.asyncio
-async def test_oauth(mocker):
+async def test_oauth_without_fastapi():
+    with pytest.raises(FastAPIImportError):
+        import spylib.oauth  # noqa
```
hillairet: Better to ignore specific error and not all of them. I think that's the right one:
```suggestion
        import spylib.oauth  # noqa: F401
```

tests/oauth/test_base.py
```diff
@@ -61,7 +61,20 @@ def json(self):
...
+
+
+@pytest.mark.asyncio
+async def test_oauth_with_fastapi(mocker):
+    if 'fastapi' not in modules and util.find_spec('fastapi') is None:
+        return
+
+    from fastapi import FastAPI  # type: ignore
+    from fastapi.testclient import TestClient  # type: ignore
```
hillairet: Here too, please ignore only the error that we have to ignore, not all of them.

hillairet (changes requested):
Minor things. Wording and syntax.

scripts/test_suite.sh
lishanl: was having skipped tests. not any more I will get rid of it

hillairet (approved):
So cool to have!

PR #103, opened 2022-03-22
Remove loguru and replace with the standard logging

lishanl:
resolve part of  #84


hillairet (approved):
So nice to have tests for this kind of things!

PR #104, opened 2022-03-28
Fix/replace shortuuid dependency with code

### LINKED ISSUE #84: Reduce dependencies

lishanl:
Fix #84 
Copy over shortuuid code used in `get_unique_id()`

@hillairet Do you have on mind how to credit to the shortuuid lib? Thanks


hillairet:
@lishanl Good question!
A comment in the code would be a good start.

@qw-in Do you know what is the best practice when grabbing code from a library like this?

hillairet:
@jsemearns I just remembered that we moved this to toolip.
We at least need toolip to also copy over the code. I don't think we want toolip as a dependency in SPyLib though

hillairet (approved):
Not much code to get rid of a dependency! Sweet!

qw-in:
@hillairet not sure - I usually just add a permalink to the code I copied. Could also put in a vendor's folder but that is likely overkill

spylib/utils/constants.py
```diff
@@ -0,0 +1 @@
+ALPHABET = '23456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'
```
qw-in: Why does the alphabet start at 2?

I'd suggest keeping all the code copied from an external library all in one file or folder

hillairet:
> @hillairet not sure - I usually just add a permalink to the code I copied. Could also put in a vendor's folder but that is likely overkill

Cool so a simple comment "Code copied over from shortuuid package: <PERMALINK>" would be fine.
Then having this at the top of the file (called shortuuid???) would be logical.

spylib/utils/constants.py
lishanl: I wondered that too, but it's just copied from the code https://github.com/skorokithakis/shortuuid/blob/master/shortuuid/main.py#L42

qw-in (approved):

hillairet (approved):
A good thing done!

I really wonder about this missing 1 in the alphabet.
Damnit @qw-in! Now I need to know!  :wink: 
@lishanl If you don't mind looking a little bit that'd be lovely but don't spend too much time on this.

lishanl:
> A good thing done!
> 
> I really wonder about this missing 1 in the alphabet. Damnit @qw-in! Now I need to know! 😉 @lishanl If you don't mind looking a little bit that'd be lovely but don't spend too much time on this.

🕵🏼‍♀️

lishanl:
> A good thing done!
> 
> I really wonder about this missing 1 in the alphabet. Damnit @qw-in! Now I need to know! 😉 @lishanl If you don't mind looking a little bit that'd be lovely but don't spend too much time on this.

I hope the answer by the Author doesn't disappoint you. 😉 
https://github.com/skorokithakis/shortuuid/pull/33#issuecomment-313862131
@hillairet @qw-in

PR #106, opened 2022-03-29
🐛 Fix/Shopify graphql error not surfacing

### LINKED ISSUE #86: Swallowing graphql errors
https://github.com/SatelCreative/spylib/blob/379793efd918273abb73010ae1d711e089d989bb/spylib/token.py#L221

This line is swallowing graphql errors. It is very common that [graphql will have an error but still return some data](https://graphql.org/learn/serving-over-http/#response).

My proposal would be to:
1. By default raise anytime errors are present
2. Have an argument or flag that can suppress errors

It is also important to surface errors back to the caller if possible

lishanl:
Fix #86 
 - add `suppress_errors` flag and make it `True` by default to make it backward compatible 
 - catch and raise general Shopify graphql error with the json response when `suppress_errors` is set to `False`


tests/token/test_graphql_errors.py
```diff
@@ -0,0 +1,31 @@
...
+    token = await OfflineToken.load(store_name=test_information.store_name)
+
+    query = """
+    {
+      something {
+        name
+      }
+    }"""
+
+    gql_response = {'data': {}, 'errors': []}
```
qw-in: Why does it raise for this response? Is that the old behaviour?

qw-in (commented):
I think the old behaviour was:
- If not data and errors then raise
- If data and errors return data

Just want to confirm that is still the default?

tests/token/test_graphql_errors.py
lishanl: Hey @qw-in , this is a new behavior when the `suppress_errors` is set to `False` to surface the errors mentioned in here https://graphql.org/learn/serving-over-http/#response you brought up in in the issue https://github.com/SatelCreative/spylib/issues/86
>a query might result in some data and some errors,

lishanl:
> I think the old behaviour was:
> 
> * If not data and errors then raise
> * If data and errors return data
> 
> Just want to confirm that is still the default?

sorry.. the code is confusing.. I will see what to do better.

tests/token/test_graphql_errors.py
qw-in: Guess I'm confused - the errors array is empty?

lishanl: oh that.. because it will raise the exception as long as there's the `errors` field, so I thought it didn't matter what's in the array. And I just added in some error message to hopefully eliminate the confusion

qw-in:
I think I'll have to play with this one a bit - will have a better look tomorrow

lishanl:
> I think I'll have to play with this one a bit - will have a better look tomorrow

sounds good, thank you @qw-in  👍🏼

hillairet (approved):
I haven't tested but looks good.

spylib/token.py
```diff
@@ -218,6 +220,10 @@ async def execute_gql(
                 'Flag the access token as invalid.'
             )
             raise ConnectionRefusedError
+
+        if not suppress_errors and 'data' in jsondata and 'errors' in jsondata:
+            raise ShopifyGQLError(jsondata)
```
qw-in: If we move this below the existing block, could we remove the `'data' in jsondata`?

Also, we likely only want to raise if the `errors` array contains at least one item...can't remember if that check achieves this

qw-in (commented):
@lishanl we talked about it during the previous meeting so just following up - I think we agree `supress_errors=False` would be a much more sane default.

cc @hillairet would you be ok to make this a breaking change?

hillairet:
> @lishanl we talked about it during the previous meeting so just following up - I think we agree `supress_errors=False` would be a much more sane default.
> 
> cc @hillairet would you be ok to make this a breaking change?

If we are encountering places where we think a better default or a code change is needed but will lead to breaking change, I think now is a better time than later.
So go for it.

qw-in (approved):
Awesome! Thanks lishan 🎉

spylib/token.py
```diff
@@ -263,6 +260,9 @@ async def execute_gql(
             else:
                 raise ValueError(f'GraphQL query is incorrect:\n{errorlist}')
 
+        if not suppress_errors and 'errors' in jsondata and len(jsondata['errors']) >= 1:
```
hillairet: Little suggestion for shorter:

```suggestion
        if not suppress_errors and len(jsondata.get('errors', [])) >= 1:
```

lishanl:
Thanks Quinn and Ant!

PR #112, opened 2022-04-14
✨ Feature/webhook create

### LINKED ISSUE #109: Add webhook create method to Token
### Discussed in https://github.com/SatelCreative/spylib/discussions/105

<div type='discussions-op-text'>

<sup>Originally posted by **qw-in** March 28, 2022</sup>
It would be really helpful if SpyLib had support for webhooks!

Here is my proposal for basic webhook support:

```python
from enum import Enum

from pydantic import BaseModel


class WebhookTopic(Enum):
    ORDERS_CREATE = 'orders/create'
    # ...


class WebhookResponse(BaseModel):
    id: str
    resourceId: str


class Token:
    # ...

    async def webhook_create(
        self,
        topic: WebhookTopic | str,
        callback_url: str,
        include_fields=None,
        metafield_namespaces=None,
        private_metafield_namespaces=None,
    ) -> WebhookResponse:
        """Uses graphql to create a webhook"""
        ...


```</div>

### PR DESCRIPTION
**Pull request by lishanl**
fix #109 

- add `webhook_http_create` for creating webhooks using [`webhookSubscriptionCreate`](https://shopify.dev/api/admin-graphql/2022-01/mutations/webhookSubscriptionCreate) with `callback_url`
- add happy path and user error tests 

**Questions:**
just checking before I proceed
- I just put the payload in the method right now. Do you have a better way to put it, e.g. in a file like we do in other repos? 
- I added the placeholders for the webhook create methods for the aws event bridge and the google cloud pubsub ones. I plan to implement them like `webhook_http_create` using [eventBridgeWebhookSubscriptionCreate](https://shopify.dev/api/admin-graphql/2022-01/mutations/eventBridgeWebhookSubscriptionCreate), [pubSubWebhookSubscriptionCreate](https://shopify.dev/api/admin-graphql/2022-01/mutations/pubSubWebhookSubscriptionCreate) and . Do you have on mind any other plan for their implementation? 
- I am thinking to copy our most used webhook topics into the enum from the Shopify list [here](https://shopify.dev/api/admin-graphql/2022-01/enums/WebhookSubscriptionTopic). OR do you suggest to just copy all?

### DIFF
--- a/spylib/token.py
+++ b/spylib/token.py
@@ -5,7 +5,7 @@ from datetime import datetime, timedelta
 from enum import Enum
 from math import ceil, floor
 from time import monotonic
-from typing import Any, ClassVar, Dict, List, Optional
+from typing import Any, ClassVar, Dict, List, Optional, Union
 
 from httpx import AsyncClient, Response
 from pydantic import BaseModel, validator
@@ -278,7 +278,7 @@ class Token(ABC, BaseModel):
 
     async def create_http_webhook(
         self,
-        topic: WebhookTopic,
+        topic: Union[WebhookTopic, str],
         callback_url: str,
         include_fields=None,
         metafield_namespaces=None,
@@ -308,7 +308,7 @@ class Token(ABC, BaseModel):
             }
         '''
         variables = {
-            'topic': topic.value,
+            'topic': topic.value if isinstance(topic, WebhookTopic) else topic,
             'webhookSubscription': {
                 'callbackUrl': callback_url,
                 'format': WEBHOOK_JSON_FORMAT,
--- a/spylib/token.py
+++ b/spylib/token.py
@@ -2,6 +2,7 @@ import logging
 from abc import ABC, abstractclassmethod, abstractmethod
 from asyncio import sleep
 from datetime import datetime, timedelta
+from enum import Enum
 from math import ceil, floor
 from time import monotonic
 from typing import Any, ClassVar, Dict, List, Optional
@@ -53,6 +54,14 @@ class OnlineTokenResponse(BaseModel):
     expires_in: int
     associated_user_scope: str
     associated_user: AssociatedUser
+
+
+class WebhookTopic(Enum):
+    ORDERS_CREATE = 'ORDERS_CREATE'
+
+
+class WebhookResponse(BaseModel):
+    id: str
 
 
 class Token(ABC, BaseModel):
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: spylib/token.py [hunk 1]
```diff
@@ -265,6 +276,76 @@ async def execute_gql(
...
+                    message
+                }
+              }
+            }
+        '''
+        variables = {
+            'topic': topic.value if isinstance(topic, WebhookTopic) else topic,
+            'webhookSubscription': {
+                'callbackUrl': callback_url,
+                'format': WEBHOOK_JSON_FORMAT,
```
qw-in: I think setting this directly in the query is probably fine

qw-in (approved):
Looks good to me! Thanks @lishanl!

lishanl:
I ran brunette with option `--skip-string-normalization` to reformat the files locally but it didn't change anything. Not sure what's wrong in the tests, do you guys know?

qw-in:
> I ran brunette with option --skip-string-normalization to reformat the files locally but it didn't change anything. Not sure what's wrong in the tests, do you guys know?

I wonder if it has issues on python 3.10 - I can look into it next week

hillairet:
Could we do a bit of scraping to get all the topics?
We will have to rerun it every time there is a new API version though ...
Maybe it's better to handle the API error for a wrong topic instead of hard coded them? :thinking:

spylib/token.py
```diff
@@ -265,6 +276,76 @@ async def execute_gql(
...
+    async def create_event_bridge_webhook(
+        self,
+        topic: Union[WebhookTopic, str],
+        arn: str,
+        include_fields=None,
+        metafield_namespaces=None,
+        private_metafield_namespaces=None,
+    ) -> WebhookResponse:
+        # TODO
+        pass
```
hillairet: I think that'd be clearer:
```suggestion
        raise NotImplementedError
```

spylib/token.py
hillairet: Yeah it's constant/hardcoded in both cases.
And if it becomes a variable later for some reason it'll need to be refactored in both cases.

tests/token/test_webhook_create.py
```diff
@@ -0,0 +1,81 @@
...
+        return_value=MockHTTPResponse(status_code=200, jsondata=gql_response),
+    )
+
+    res = await token.create_http_webhook(
+        topic=WebhookTopic.ORDERS_CREATE,
+        callback_url='https://example.org/endpoint',
+        include_fields=["id", "note"],
+    )
+
+    shopify_request_mock.assert_called_once()
```
hillairet, line 46: I guess another possible test would be to check the values in the variables sent with the graphql mutation ... Just a suggestion. That could be done in a future PR.

hillairet (approved):

lishanl:
> > I ran brunette with option --skip-string-normalization to reformat the files locally but it didn't change anything. Not sure what's wrong in the tests, do you guys know?
> 
> I wonder if it has issues on python 3.10 - I can look into it next week

This is resolved by adding `--skip-string-normalization` to the tox brunette command
https://github.com/SatelCreative/spylib/blob/feature/webhook-create/tox.ini#L40
@hillairet @qw-in any concern about this?

tests/token/test_webhook_create.py
lishanl: I added a test, is that along the line with what you thought of?

qw-in:
@lishanl I'll migrate us to blue asap. It seems more stable - so no concerns

PR #113, opened 2022-04-19
✨ 💥 Feature/webhook verification

### LINKED ISSUE #110: Add function to verify specifically the webhook's HMAC
Following up on discussion #105 we should have a simple function to verify the webhook and it should be able to just call existing code.

```python
# Simple but a great building block
# https://shopify.dev/apps/webhooks/configuration/https#step-5-verify-the-webhook -> click python
def verify_webhook(data: str, hmac_header: str, api_secret_key: str) -> bool:
    ...
```

lishanl:
fix #110 

- rename hmac functions param `is_base64` to `use_base64`
- add ~~`verify_webhook`~~ `is_webhook_valid` function
- add tests

### DIFF
--- a/spylib/token.py
+++ b/spylib/token.py
@@ -30,6 +30,7 @@ from spylib.exceptions import (
     ShopifyThrottledError,
     not_our_fault,
 )
+from spylib.utils.hmac import validate
 from spylib.utils.rest import Request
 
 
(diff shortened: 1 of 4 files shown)


spylib/token.py
```diff
@@ -400,3 +401,11 @@ def load(cls, store_name: str):
         therefore the developer should override this.
         """
         pass
+
+
+def verify_webhook(data: str, hmac_header: str, api_secret_key: str) -> bool:
```
hillairet: If you return a boolean, I think a better name would be this:
```suggestion
def is_webhook_valid(data: str, hmac_header: str, api_secret_key: str) -> bool:
```

This way one can have a readable if statement:
```python
if is_webhook_valid():
```

hillairet (changes requested):
Because you changed the arguments of a couple of functions, you should probably use a breaking change emoji for the PR title.

hillairet (approved):
Cool stuff!

qw-in (approved):
Awesome thanks @lishanl! Excited to rip out our custom webhook code

PR #122, opened 2022-04-25
📝 Docs/syntax highlighting

### LINKED ISSUE #97: Website syntax highlighting
Currently, syntax highlighting does not appear to be working on the documentation website:

https://satelcreative.github.io/spylib/#implement-token-classes

lishanl:
fix #97 

enable mkdoc.yml syntax highlighting
```yml
markdown_extensions:
- pymdownx.superfences
```
which handles [code highlighting](https://facelessuser.github.io/pymdown-extensions/extensions/superfences/#code-highlighting)  

syntax_highlighting

Do we still want to use `max-include` which allows referencing the code samples from individual `.py` files? Do we want to organize the sample codes into files like fastapi repo does?


hillairet (approved):
Sweet! The goal of the issue was to have syntax highlighting. This PR does that so that's good enough for now.

I think the interest of having the code in a separate `.py` file is to be able to insert the same code multiple times, for example to highlight and discuss different lines from the same snippet.
[Fastapi example](https://fastapi.tiangolo.com/tutorial/middleware/)

When we need it we'll do it.

PR #126, opened 2022-05-04
✨ Add event bridge and pub sub webhook creation 

### LINKED ISSUE #116: Add handling of eventbridge to webhook creation
Implement [this function](https://github.com/SatelCreative/spylib/blob/00065a4c8db3243fd5438edfde343b1a731c3cb2/spylib/token.py#L326), its test(s) and its corresponding documentation.

### LINKED ISSUE #117: Add handling of google's pub/sub to webhook creation
Implement [this function](https://github.com/SatelCreative/spylib/blob/00065a4c8db3243fd5438edfde343b1a731c3cb2/spylib/token.py#L337), its test(s) and its corresponding documentation.

lishanl:
fix #116 
fix #117 

- move the webhook graphql implementation details to webhook utils
- add event bridge webhook creation function
- add pub sub webhook creation function
- add tests
- add doc


hillairet (changes requested):
Still in discussion but might as well go for something that looks better! Please move `/utils/webhook.py` to `/webhook.py`.

Do not generate graphql queries!
Make a hardcoded file `.graphql` with the hardcoded operation name and pass the inputs as `variables` to the GQL. I have examples if you need them.

lishanl:
> Still in discussion but might as well go for something that looks better! Please move `/utils/webhook.py` to `/webhook.py`.
> 
> Do not generate graphql queries! Make a hardcoded file `.graphql` with the hardcoded operation name and pass the inputs as `variables` to the GQL. I have examples if you need them.

I agree by putting the gql queries into files and read from them would be easier to follow (will provide easy and better readability). It can probably reduce lines of codes overall (no need tests for the generate functions). I do have one concern since the reading from file operation is a file I/O operation, would it have impact to overall performance when it comes to a high volume of concurrent requests? (though not for webhooks but for other future ones since this would be the direction we are going with).
Maybe I am over thinking it? @hillairet

hillairet:
Fair concern @lishanl.
Read the GraphQL file at the route of the module so that it is done once when the app is loaded.
If this is really a concern, the graphql should be a hardcoded string nonetheless, not a dynamic one.

spylib/utils/misc.py
```diff
@@ -9,3 +9,8 @@ def now_epoch() -> int:
 
 def get_unique_id() -> str:
     return random(length=10)
+
+
+def snake_to_camel_case(text: str) -> str:
+    temp = text.lower().split('_')
+    return temp[0] + ''.join(element.title() for element in temp[1:])
```
qw-in: Where is this from? How do we know it is correct?

spylib/utils/webhook.py
```diff
@@ -0,0 +1,91 @@
...
+    webhook_subscription_input: WebhookSubscriptionInput,
+) -> str:
+    """Generate the webhook creation graphql query"""
+    operation = webhook_subscription_create.value
+    input_type = webhook_subscription_input.value
+
+    query = f'''
+    mutation {operation}($topic: WebhookSubscriptionTopic!,
+                                        $webhookSubscription: {input_type}) {{
+        {operation}(topic: $topic, webhookSubscription: $webhookSubscription) {{
```
qw-in: Why are we doing string manipulation? Generally bad practice for GraphQL queries. Consider using a fragment instead for code reuse

spylib/utils/webhook.py
```diff
@@ -0,0 +1,91 @@
...
+    topic: Union[WebhookTopic, str],
+    include_fields: Optional[List[str]] = None,
+    metafield_namespaces: Optional[List[str]] = None,
+    private_metafield_namespaces: Optional[List[str]] = None,
+    **endpoint_kwargs,
+) -> dict:
+    """Generate the webhook creation variables"""
+    endpoint_kwargs_camel_case_name = {}
+    for key, value in endpoint_kwargs.items():
+        endpoint_kwargs_camel_case_name[snake_to_camel_case(text=key)] = value
```
qw-in: I don't love this pattern. I assume this is done to reduce duplicate code however it comes at the cost of typing & being explicit. I would prefer something like the following even if it is longer:

```python
def generate_shared_variables(
    topic: Union[WebhookTopic, str],
    include_fields: Optional[List[str]] = None,
    metafield_namespaces: Optional[List[str]] = None,
    private_metafield_namespaces: Optional[List[str]] = None,
):
    pass
  

# Do this in the method where is it needed
def generate_http_variables(..., url: str):
    generate_shared_variables()
    # merge vars here
```

Unfortunately [ParamSpec](https://docs.python.org/3/library/typing.html#typing.ParamSpec) is a `3.10` feature but one day!

qw-in (changes requested):

spylib/utils/webhook.py
qw-in: Oh, @hillairet already covered this

spylib/utils/webhook.py
lishanl: Thanks Quinn. I will change to the code to load the qgl like Ant suggested as well.

spylib/utils/misc.py
lishanl: some code I came up with, there are tests to it. Anyhow this won't be needed after the change

qw-in: Oh sorry I missed the tests! Sounds good

hillairet (commented):
Should we merge this first and then reorganize the code according to the discussion #127 ? (It hasn't received any down vote :wink: )
This way this PR keeps its focus.

lishanl:
> Should we merge this first and then reorganize the code according to the discussion #127 ? (It hasn't received any down vote 😉 ) This way this PR keeps its focus.

I am fine with that.

spylib/token.py
```diff
@@ -331,20 +310,58 @@ async def create_event_bridge_webhook(
...
+            },
+        }
+        res = await self.execute_gql(
+            query=WEBHOOK_CREATE_GQL,
+            operation_name=WebhookSubscriptionCreate.EVENT_BRIDGE.value,
+            variables=variables,
+        )
+        webhook_create = res.get(WebhookSubscriptionCreate.EVENT_BRIDGE.value, None)
+        if webhook_create and webhook_create.get('userErrors', None):
+            raise ShopifyGQLUserError(res)
```
hillairet, line 332: Isn't this check already done in `execute_gql`?

hillairet (approved):
Looks good.
One thing I'd like to do is shrink "WebhookSubscriptionCreate" to "WebhookCreate" everywhere cause it's really too long otherwise.
This could be done when the code is reorganized if you prefer.

spylib/webhook.py
```diff
@@ -0,0 +1,20 @@
+from enum import Enum
+from pathlib import Path
+
+from pydantic import BaseModel
+
+WEBHOOK_CREATE_GQL = Path('./spylib/webhook.graphql').read_text()
```
qw-in, line 6: I assume the `.graphql` won't get stripped out when deploying to something like an aws lambda?

hillairet: Fair question but I'd be really surprised if it sifted through the files at deployment to allow only specific extensions!
It may have some protections against executable files but not a plain text file.
It would be nice to double check though.

lishanl: Did we encounter some problem like that before? 

> I assume the `.graphql` won't get stripped out when deploying to something like an aws lambda?

lishanl: Alternatively, we can make it a string.

hillairet: I have never seen this as a problem. Let's keep like this until we have a reason to change it back to string.

spylib/utils/webhook.py
lishanl: Good to know about the [ParamSpec](https://docs.python.org/3/library/typing.html#typing.ParamSpec).
Yea, the intention was to stay DRY (do not repeat yourself), avoid copy and paste.  Thanks for pointing out the typing issue, @qw-in . I thought about making the input with pydantic basemodel. what would you think of that
```py
class Webhook(BaseModel):  
    topic: Union[WebhookTopic, str] 
    includeFields: Optional[List[str]] = None
    metafieldNamespaces: Optional[List[str]] = None
    privateMetafieldNamespaces: Optional[List[str]] = None

class HTTPWebhook(Webhook):
    callbackUrl: str

class EventBridgeWebhook(Webhook):
    arn: str

class PubSubWebhook(Webhook):
    ...



```

spylib/webhook.py
lishanl: ok.

qw-in (approved):

PR #133, opened 2022-05-12
:technologist: Brunette => Blue

qw-in:
Blue supports 3.10 and has generally been a better experience. Also I found some `"`


lishanl (approved):
cool, I like their branding too.

PR #134, opened 2022-05-12
👷 fix dependencies reinstallation

lishanl:
Quinn pointed out that the dependencies have been reinstalled many times in the python 310 test environment and the overall CI takes a long time to run
tests

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,7 @@ python =
 [testenv]
 whitelist_externals = poetry
 commands =
+    poetry install -v
     poetry run pytest --cov=spylib --cov-report term-missing tests/
 
 [testenv:mypy]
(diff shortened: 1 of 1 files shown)


lishanl:
This change has shortened the total run time from 12 min to 5 min by running all tests and checks in each Python version because it doesn't create an isolated test env and install all dependencies again for each check such as mypy, ...  At this point installing dependencies take longer than running the tests and checks

Can you think of a better optimization?

qw-in (approved):
Awesome start!

lishanl:
> Awesome start!

Have to say that long running time has been a pain. Thanks for looking into it and find what's causing it @qw-in

hillairet (approved):
I can see that the processing time for these tests would be an issue. Now we have another one though: If flake8 fails, then the other tests don't run, correct?

I guess it's good enough to get the user to test locally and fix all the issues before committing again.
If we can have the best of both worlds though, that'd be awesome!

PR #137, opened 2022-05-18
👽️ 🔥 Fix/remove jwt

### LINKED ISSUE #131: Remove all the JWT for cookie in iframe
Since cookies in iframes is already not supported by some browser and soon by all of them.
Instead of using cookies to authenticate the backend calls for an app's frontend, one should use Shopify's session tokens.

So let's get rid of all the `jwtoken` stuff.

lishanl:
fix #131 

my understanding is that we only remove the jwtoken used to set cookie with not the jwtoken used for the oauth [state/{nonce}](https://shopify.dev/apps/auth/oauth/getting-started#step-3-ask-for-permission) param.

### DIFF
--- a/scripts/test_suite.sh
+++ b/scripts/test_suite.sh
@@ -23,7 +23,7 @@ FLAKE8OUT=`flake8`
 reportvalidation "$FLAKE8OUT"
 
 echo -ne "\n######### CHECK FORMATTING: "
-BLACKOUT=`brunette spylib tests --check 2>&1 --skip-string-normalization`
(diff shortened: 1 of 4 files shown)


hillairet (approved):
Nice to have this clean up done!

lishanl:
@hillairet do we know of any app that's using the cookie thing from here? Just wonder if we need to flag this as a breaking change. thanks

hillairet:
Yeah that's a breaking change. I don't think it's used though.

PR #142, opened 2022-05-28
♻️💥 Fix/reorganize modules

### LINKED ISSUE #129: Reorganize module structure to be more consistent
The current file/module organization lacks coherence at this time.
It's not clear why `session_token.py` belongs in "utils" for example.

I think we have a natural organization for the files that would match the documentation menu to highlight the Shopify features handled by SPyLib:
```
spylib
├── admin_api.py       # Instead of tokens.py
├── hmac.py            # Exposing to the user more directly. Maybe should be added to docs menu.
├── oauth
├── session_token.py
├── webhook.py
```

This way we have a natural organization for future Shopify features that SPyLib handles and it makes the imports a lot more logical:
```python
from spylib.session_token import SessionToken
```

Also we should reorganize the webhooks into functions to avoid making the `Token` class bloated. We can simply have:
```python
from spylib import webhook

webhook_response = webhook.create_http(
    offlinetoken=offlinetoken,
    topics=['orders/create'],
    callback_url=`https://example.com/webhooks
)
```

_Originally posted by @hillairet in https://github.com/SatelCreative/spylib/discussions/127_

Additionally let's not expose the functions and classes at the top level of the package.
In other words this should not be possible:
```python
from spylib import OfflineTokenABC
```

Instead we can only do:
```python
from spylib.admin_api import OfflineTokenABC
```

lishanl:
fix #129 

- move `session_token`, `hmac` modules to top level
- add doc for hmac
- rename `token.py` to `admin_api.py`
- remove direct access to classes and functions from top level

I will do the webhook part in a separate PR


hillairet (changes requested):
Starting to look a lot more organized!
Could you reorganize the tests folders and filenames to match the new structure of the code?

Incomplete list:
* hmac and session token tests out of the `/tests/utils` folder
* `/tests/token` -> `/tests/admin_api`

hillairet (approved):

PR #144, opened 2022-05-31
:sparkles: auth utility functions

qw-in:
I outlined how I see the utility functions being implemented by partially implementing the `token_exchange` step of oauth. Please provide some feedback to I can move ahead with this - I have to implement it for an app anyway so would like to do it here

Note: I think I found a bug - you cannot import anything from the oauth folder without having fastapi installed because of the import in `__init__`

### DIFF
--- a/spylib/oauth/perform_token_exchange.py
+++ b/spylib/oauth/perform_token_exchange.py
@@ -1,0 +1,116 @@
+from typing import List, Union
+
+from httpx import AsyncClient, codes
+from pydantic import BaseModel, validator
+
+
+def parse_scope(scope: str) -> List[str]:
+    return scope.split(',')
+
+
+# Docstrings would be fully completed
+# Types would potentially be moved to .types
+
+
+class OfflineTokenResponse(BaseModel):   # I don't love this naming
+    """
+    [Read more about Offline access](https://shopify.dev/apps/auth/oauth/access-modes#offline-access)
+    """
+
+    access_token: str
+    """
+    An API access token that can be used to access the shop’s data as long as your app is installed. Your app should store the token somewhere to make authenticated requests for a shop’s data.
+    """
+    scope: List[str]
+    """
+    The list of access scopes that were granted to your app and are associated with the access token.
+    """
+
+    _normalize_scope = validator('scope', allow_reuse=True, pre=True)(parse_scope)
+
+
+class AssociatedUser(BaseModel):
+    id: int
+    first_name: str
+    last_name: str
+    email: str
+    email_verified: bool
+    account_owner: bool
+    locale: str
+    collaborator: bool
+
+
+class OnlineTokenResponse(BaseModel):
+    """
+    [Read more about Online access](https://shopify.dev/apps/auth/oauth/access-modes#online-access)
+    """
+
+    access_token: str
+    """
+    
+    """
+    scope: List[str]
+    """
+    The list of access scopes that were requested to be associated with the access token.
+    """
+
+    _normalize_scope = validator('scope', allow_reuse=True, pre=True)(parse_scope)
+
+    expires_in: int
+    associated_user_scope: List[str]
+
+    _normalize_associated_user_scope = validator(
+        'associated_user_scope', allow_reuse=True, pre=True
+    )(parse_scope)
+
+    """
+    The list of access scopes that were granted to associated with the access token.
+    """
+    associated_user: AssociatedUser
+
+
+async def perform_token_exchange(
+    *,
+    code: str,
+    shop: str,
+    api_key: str,
+    api_secret_key: str,
+) -> Union[OnlineTokenResponse, OfflineTokenResponse]:
+    """_summary_
+
+    Args:
+        code (str): _description_
+        shop (str): _description_
+        api_key (str): _description_
+        api_secret_key (str): _description_
+
+    Raises:
+        Exception: _description_
+
+    Returns:
+        Union[OnlineTokenResponse, OfflineTokenResponse]: _description_
+    """
+
+    async with AsyncClient(base_url=f'https://{shop}') as client:
+        response = await client.post(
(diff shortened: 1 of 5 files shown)


spylib/oauth/perform_token_exchange.py
```diff
@@ -0,0 +1,116 @@
...
+class OnlineTokenResponse(BaseModel):
+    """
+    [Read more about Online access](https://shopify.dev/apps/auth/oauth/access-modes#online-access)
+    """
+
+    access_token: str
+    """
+    
+    """
+    scope: List[str]
```
qw-in, line 52: These _look_ like duplicates but would be documented differently which is why there is duplication here

qw-in:
Oh, this looks fantastic https://lundberg.github.io/respx/

hillairet:
Hummmmm ... I'm kinda confused. What's wrong with [this function](https://github.com/SatelCreative/spylib/blob/59e87bce74b6bae240f9a72be591a4e4c6fe2fcc/spylib/oauth/tokens.py#L15)?
I'm ok with renaming, changing the API, modifying/improving the function but why start from scratch?
Also I agree the class names are wonky ... So hard to find something satisfying.

Definitely I wouldn't have a function returning `Union[OfflineTokenResponse, OnlineTokenResponse]` but instead have two functions `exchange_online_token` and `exchange_offline_token` and those would call a third function that returns just the JSON. It's like what we did already [here](https://github.com/SatelCreative/spylib/blob/59e87bce74b6bae240f9a72be591a4e4c6fe2fcc/spylib/oauth/tokens.py#L50) but in function instead of classmethods ... I think classmethods make sense for this but I'm not set on it.

Also I'd stay clear of `perform_` which really doesn't say much. Especially when it's `perform_token_exchange`. Can we not say `exchange_token`? Shorter and starting with a meaningful verb.

Oh and yeah respx looks great!

qw-in:
> Hummmmm ... I'm kinda confused. What's wrong with this function?

Nothing. They are very different but in the actual implementation we could modify the existing function instead I suppose. I don't really see the benefit to that.

The big difference is that this would be in the public api of spylib. This and functions like it

> Definitely I wouldn't have a function returning Union[OfflineTokenResponse, OnlineTokenResponse]

Why not? That is the response type of this "api call" if you want to think of it that way?

Some implementors may not want to set different callback endpoints or use the nonce to distinguish between offline & online tokens

I do see your point though, it could be annoying for an offline only app. I really don't think returning untyped json/dict is a good solution though...

> I'd stay clear of perform_ which really doesn't say much. Especially when it's perform_token_exchange. Can we not say exchange_token? Shorter and starting with a meaningful verb.

Yeah, I agree the name needs work. I like `exchange_token` but feels like it is missing some context. Maybe `exchange_temporary_token`?

`exchange_code_for_token`?

hillairet:
Ok let me clarify what I meant instead of your function returning 2 types:
```python
async def exchange_json_token(...) -> dict:
async def exchange_offline_token(...) -> OfflineTokenResponse:
async def exchange_online_token(...) -> OnlineTokenResponse:
```

I feel like the code should know if it's doing offline or online token retrieval.

I don't understand the "temporary" in `exchange_temporary_token`.
Remember that one can do:
```python
from spylib import oauth

oauth.exchange_token(...)
```
So it does have context from the `oauth` part.
Why isn't it `retrieve_token` even? What is exchanged?

qw-in:
Would you be willing to meet me in the middle with this?

```python
async def exchange_json_token(...) -> OnlineTokenResponse | OfflineTokenResponse:

async def exchange_offline_token(...) -> OfflineTokenResponse:

async def exchange_online_token(...) -> OnlineTokenResponse:
```

---

I believe this step is actually called exchange in oauth. The temporary token is the `code` passed in as a query param

That said I like retrieve, I'd agree on `retrieve_token`. Will have docs anyway

lishanl:
`exchange...` sounds fine to me. Shopify does also say "exchange" in their doc https://shopify.dev/apps/auth/oauth/getting-started#step-5-get-a-permanent-access-token
>If all security checks pass, then you can exchange the access code for a permanent access token by sending a request to the shop’s access_token endpoint

Returning specific types sounds good. 
```py
async def exchange_json_token(...) -> Union[OnlineTokenResponse, OfflineTokenResponse]:

async def exchange_offline_token(...) -> OfflineTokenResponse:

async def exchange_online_token(...) -> OnlineTokenResponse:
```

Are we 
 - modifying the existing `oauth/tokens.py`?
 or 
 - adding in the new `token_exchange.py` module and deprecating the `tokens.py`?

hillairet:
> `exchange...` sounds fine to me. Shopify does also say "exchange" in their doc https://shopify.dev/apps/auth/oauth/getting-started#step-5-get-a-permanent-access-token
> 
> > If all security checks pass, then you can exchange the access code for a permanent access token by sending a request to the shop’s access_token endpoint

Oh ok! I didn't realize that the step was called exchange.
The docstring could link to that specific section!


> ```python
> async def exchange_token(...) -> Union[OnlineTokenResponse, OfflineTokenResponse]:
> 
> async def exchange_offline_token(...) -> OfflineTokenResponse:
> 
> async def exchange_online_token(...) -> OnlineTokenResponse:
> ```

:point_up: I just renamed the first one cause "json" doesn't make sense.
That's a fair compromise. Offers flexibility and clarify IMO.

> Are we
> 
> * modifying the existing `oauth/tokens.py`?
>   or
> * adding in the new `token_exchange.py` module and deprecating the `tokens.py`?

I'd say keep the classes in `tokens.py` and the `exchange_token` function could be in `token_exchange.py` or `exchange_token.py` or something else.

Also do something with the current `_get_token`. It's not broken so it has some valid code. Still not sure why we restart from scratch for that.

qw-in:
Closing in favor of #146

Issue #132, opened 2022-05-12
Merge the OnlineToken/OfflineToken response classes


hillairet:
Currently we have 2 OfflineToken/OnlineToken response classes:
* [In oauth](https://github.com/SatelCreative/spylib/blob/98154d88121347db475d4b6e922636e467ef4d3a/spylib/oauth/tokens.py#L45)
* [In token](https://github.com/SatelCreative/spylib/blob/98154d88121347db475d4b6e922636e467ef4d3a/spylib/token.py#L48)

Let's move the classes `AssociatedUser` `OfflineTokenResponse` `OnlineTokenResponse` from `spylib/token.py` to `spylib/models.py`.
Then all of the oauth code should use those :point_up: 

While at it, let's extract get rid of the method (https://github.com/SatelCreative/spylib/blob/98154d88121347db475d4b6e922636e467ef4d3a/spylib/oauth/tokens.py#L50) and instead have the functions `fetch_online_token` and `fetch_offline_token` which will simply call `_get_token`, in the same file as now.

PR #146, opened 2022-06-02
:sparkles:💥 [oauth] add exchange_token

qw-in:
Main changes:
- added the `oauth.exchange_token` (and variants) - see #144 for extensive discussion
- added `OnlineTokenModel` & `OfflineTokenModel` with docstrings
- added test for `exchange_token`
- removed the old `OnlineToken` & `OfflineToken`
- updated fastapi implementation to use the new `exchange_token`
- fastapi no longer needs to be installed to import from the `oauth` folder

Unrelated changes (can move to their own pr):
Disable flake8 E501. It was preventing me from adding links to docstrings. Tried:
    - Splitting into multiple lines, but markdown was no longer parsed
    - Moving link to bottom of md github style, no longer parsed
Since blue already automatically handles line length, hoping this is not too contentious

I can add website documentation to this pr as well. Just let me know where and in what format. Otherwise could be added in a later pr


spylib/oauth/types.py
```diff
@@ -0,0 +1,111 @@
...
+    email: str
+    """
+    The user's email.
+    """
+
+    email_verified: bool = False
+    """
+    If the user's email has been verified.
+
+    If `False` or `None` email cannot be trused (legacy accounts).
```
qw-in:
```suggestion
    If `False` or `None` email cannot be trusted (legacy accounts).
```

spylib/oauth/types.py
```diff
@@ -0,0 +1,111 @@
...
+    If the user is a partner who collaborates with the merchant.
+    """
+
+
+class OnlineTokenModel(BaseModel):
+    """
+    [Read more about Online access](https://shopify.dev/apps/auth/oauth/access-modes#online-access)
+    """
+
+    access_token: str
```
qw-in: Should we use the pydantic secret string here?

Edit: I think perhaps forcing `.get_secret_value()` would be too confusing

https://pydantic-docs.helpmanual.io/usage/types/#secret-types

spylib/oauth/exchange_token.py
```diff
@@ -0,0 +1,114 @@
+from typing import Union
+
+from httpx import AsyncClient, codes
+
+from .types import OfflineTokenModel, OnlineTokenModel
+
+
+async def exchange_token(
+    *,
+    shop: str,
```
lishanl, line 10: We use the variable name `shop` to mean the store handle in [here](https://github.com/SatelCreative/spylib/blob/main/spylib/utils/domain.py)
so we probably want to consistent on that. Would it be more explicit to name it `store_domain` here ?

spylib/oauth/exchange_token.py
```diff
@@ -0,0 +1,114 @@
...
+    Can be used instead of `exchange_token` when you know by some mechanism it will only be an offline token.
+
+    [All prior security checks must already have been completed.](https://shopify.dev/apps/auth/oauth/getting-started#step-4-confirm-installation)
+
+    Args:
+        shop (str): The name of the merchant's shop. `example.myshopify.com`
+        code (str): The authorization code provided in the redirect.
+        api_key (str): The API key for the app, as defined in the Shopify Partner Dashboard.
+        api_secret_key (str): The API secret key for the app, as defined in the Shopify Partner Dashboard.
+
```
lishanl, line 73: Are the docstrings the same for `exchange_token`, `exchange_online_token`, `exchange_offline_token` except for the Returns part? If so, would we consider doing something shown [here](https://stackoverflow.com/questions/2575959/repetitive-content-in-docstrings) to avoid the repetition or that's kinda ugly?

spylib/oauth/types.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    The user's email.
+    """
+
+    email_verified: bool = False
+    """
+    If the user's email has been verified.
+
+    If `False` or `None` email cannot be trusted (legacy accounts).
+    """
```
hillairet: Comment says that `None` is a possibility but the type doesn't match:
```suggestion
    email_verified: Optional[bool] = False
    """
    If the user's email has been verified.

    If `False` or `None` email cannot be trusted (legacy accounts).
    """
```

spylib/oauth/types.py
```diff
@@ -0,0 +1,111 @@
...
+    The user's first name.
+    """
+
+    last_name: str
+    """
+    The user's last name.
+    """
+
+    email: str
```
hillairet: Similar to so `SecretStr`, we could go all the way with pydantic:
```suggestion
    email: EmailStr
```

tests/oauth/test_exchange_token.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    result = await exchange_token(
+        shop=shop,
+        code=code,
+        api_key=api_key,
+        api_secret_key=api_secret_key,
+    )
+
+    assert result == token_model.parse_obj(token_dict)
```
hillairet, line 61: I think this needs to be tested manually because both dicts go through the same model processing.
```suggestion
    assert result == token_model.parse_obj(token_dict)
    parsed_scopes = token_dict['scope'].split(',')
    assert result.scopes == parsed_scopes
    if 'associated_user_scope' in token_dict:
        parsed_user_scopes = token_dict['associated_user_scope'].split(',')
        assert result.associated_user_scopes == parsed_user_scopes
```

tests/oauth/test_exchange_token.py
```diff
@@ -0,0 +1,60 @@
+import pytest
+from respx import MockRouter
+
+from spylib.oauth import OfflineTokenModel, OnlineTokenModel, exchange_token
+
+exchange_token_params = [
+    (
+        dict(
+            access_token='offline-token',
+            scope='read_products',
```
hillairet:
```suggestion
            scope='write_products,read_orders',
```

spylib/oauth/types.py
```diff
@@ -0,0 +1,111 @@
+from typing import List
+
+from pydantic import BaseModel, validator
+
+
+def _parse_scope(scope: str) -> List[str]:
+    return scope.split(',')
+
+
+class OfflineTokenModel(BaseModel):
```
lishanl: Cool, moving the models to its own place, this partially solve https://github.com/SatelCreative/spylib/issues/132. Instead of putting them in `spylib/models.py`, following this pattern, we would have `types.py` for each package/interface I imagine. 
Do we want to merge the models from [`admin_api`](https://github.com/SatelCreative/spylib/blob/main/spylib/admin_api.py#L42-L63) in this PR as well?

tests/oauth/test_exchange_token.py
```diff
@@ -0,0 +1,60 @@
...
+        ),
+        OfflineTokenModel,
+    ),
+    (
+        dict(
+            access_token='online-token',
+            scope='write_products',
+            expires_in=86399,
+            associated_user_scope='read_products',
```
hillairet:
```suggestion
            associated_user_scope='read_products,read_orders',
```

tests/oauth/test_exchange_token.py
```diff
@@ -0,0 +1,60 @@
...
+        dict(
+            access_token='offline-token',
+            scope='read_products',
+        ),
+        OfflineTokenModel,
+    ),
+    (
+        dict(
+            access_token='online-token',
+            scope='write_products',
```
hillairet:
```suggestion
            scope='write_products,read_orders',
```

lishanl (commented):
Awesome docs. Thanks @qw-in

hillairet (changes requested):
Cool. Small tweaks

spylib/oauth/types.py
lishanl: Nice discovery. This is like a safeguard. In case, the user logs the `access_token` by accident, it masks it. I would say why not.

spylib/oauth/types.py
hillairet: I think I'd be more for `models` but not a deal breaker. Maybe @lishanl needs to be the tie breaker on naming!  :wink:

spylib/oauth/types.py
hillairet: This is something that we need to answer in general for this library.
I was gonna say that we should use `SecretStr` to use all the goodies from pydantic but then I realized that `.dict` doesn't return the string ... that seems annoying.
I'm on the fence.  :persevere:

spylib/oauth/exchange_token.py
qw-in: I expect that would cause more harm than good - especially when vscode is trying to pick up autocomplete or we are trying to generate docs. No perfect solution I'm afraid

spylib/oauth/exchange_token.py
qw-in: I expect that would cause more harm than good - especially when vscode is trying to pick up autocomplete or we are trying to generate docs. No perfect solution I'm afraid

spylib/oauth/types.py
qw-in: I couldn't decide between the two actually. Settled on types because I was planning to add some non-model types as well. I don't have a strong feeling on this one

spylib/oauth/types.py
qw-in: Hmm, so probably a documentation issue. It can be none in very niche cases on very old accounts but I'm not sure we want to represent that in the model?

I was hoping with the default false the optional could be avoided

spylib/oauth/exchange_token.py
qw-in: I'm using shop because that is consistent with all of the Shopify api calls (auth, session, webhook, etc.)

I guess I feel like we'd need a very good reason to change it?

spylib/oauth/types.py
qw-in: I'm on the fence as well. Let's see how we all feel tomorrow

Or, we could add it and see how annoying it is haha

spylib/oauth/types.py
qw-in: > Do we want to merge the models from [admin_api](https://github.com/SatelCreative/spylib/blob/main/spylib/admin_api.py#L42-L63) in this PR as well?

Can we do that in a different PR? I'm not sure why there are oauth models in the admin api

tests/oauth/test_exchange_token.py
qw-in: > I think this needs to be tested manually because both dicts go through the same model processing.

I don't really understand...do you mean in order to test the models? I can write parsing tests for those elsewhere if that is a concern

spylib/oauth/types.py
qw-in: This requires an extra package so I'm going to say lets open an issue or discussion and talk about it but not for this pr

spylib/oauth/types.py
qw-in: I've decided not for this PR

tests/oauth/test_exchange_token.py
```diff
@@ -0,0 +1,60 @@
...
+    shop = 'example.myshopify.com'
+    code = 'code'
+    api_key = 'api-key'
+    api_secret_key = 'api-secret-key'
+
+    respx_mock.post(
+        f'https://{shop}/admin/oauth/access_token',
+        json__code=code,
+        json__client_id=api_key,
+        json__client_secret=api_secret_key,
```
lishanl, line 49: Is this equivalent to the dict version below, any difference in functionality?
```py
respx_mock.post(
        f'https://{shop}/admin/oauth/access_token',
        json={ 'code': code, 'client_id': api_key, 'client_secret': api_secret_key } 
       ).respond(
        json=token_dict,
    )

```
Can we include a comment with the link to https://lundberg.github.io/respx/api/#json for quick look up since it's not a commonly used syntax?

tests/oauth/test_exchange_token.py
lishanl: yea, what do you mean "both dicts go through the same model processing" @hillairet ? `token_model` is of type `OfflineTokenModel` in first test and `OnlineTokenModel` in the second test.

spylib/oauth/types.py
lishanl: > I couldn't decide between the two actually. Settled on types because I was planning to add some non-model types as well. I don't have a strong feeling on this one



> I think I'd be more for `models` but not a deal breaker. Maybe @lishanl needs to be the tie breaker on naming! 😉

I am more for `models` in that case. 😄

spylib/oauth/exchange_token.py
lishanl: ok. I see what you mean. ref: https://shopify.dev/apps/auth/oauth/getting-started#step-4-confirm-installation

lishanl (approved):
Thanks Quinn. Please see my comments on
1. change `types.py` -> `models.py`
2. include a comment with the ref link for  the `respx`  `json__` syntax if not using the `dict` version

spylib/oauth/types.py
qw-in: I'll change it

tests/oauth/test_exchange_token.py
qw-in: @lishanl I'll add the doc link, good idea

There is a small difference I think. When passing the `json={}` the whole dict has to match exactly. With `json__` it only cares about that specific key. I guess I also like the `json__` as I don't have to use quotes - feels cleaner I guess

tests/oauth/test_exchange_token.py
hillairet: I mean if you screw up the scopes parsing, your test only won't catch it since the same dict will go through the model. It's just to test your validator essentially. One could make the argument that it should be done separately though thinking about it.

hillairet (approved):
My remaining comment on the test is not that critical

PR #147, opened 2022-06-09
:sparkles: [oauth] Add oauth hmac signature validation function

qw-in:
Add a utility function to verify signature of requests. Please help with the naming!

Why a new function:
- Specific to oauth
- Support the `ids` special case
- Add specific test cases
- Documentation


spylib/oauth/signature_validation.py
```diff
@@ -0,0 +1,81 @@
...
+    """Contains the parsed query string and the signature if provided."""
+
+    query_string: str = ''
+    """The query string parsed and stripped of the `hmac` parameter."""
+
+    signature: Optional[str] = None
+    """The value of the `hmac` query parameter if it exists."""
+
+
+def parse_signed_query_string(query_string: str) -> SignedQueryString:
```
hillairet: How about converting this to a class method of SignedQueryString? Seems more organized.
Then you simplify the naming:
```python
parsed = SignedQueryString.parse(query_string)
```

hillairet (changes requested):
Since this is part of a refactoring effort it would be good each time to not just add/rewrite code but also remove and/or integrate with existing code.
In that case it seems that this new function should be called [here](https://github.com/SatelCreative/spylib/blob/a27d1ea99d2ad65b6ff272aa26397ce2b211130f/spylib/oauth/validations.py#L21)
Maybe we'll reorganize the `validate_callback` code later on but at least we don't have the same process/validation implemented twice with different performance (the existing one doesn't handle the `ids` case but yours does).

qw-in:
@hillairet @lishanl coming back to this, I am actually wondering if exporting anything other than `validate_*` is a mistake?

Maybe the parse fn could simply be merged into the validate one? That way we lower the API surface area? Let me know what you think!

hillairet:
Another point we need to clarify for the library as a whole.
I'd say KISS, so we expose less first because it's simpler then we expose if needed. Removing/changing API causes pain to the user and docs annoyance for us so better to have less but more stability.
We can expose more once we have reached some stability.

spylib/oauth/signature_validation.py
```diff
@@ -0,0 +1,81 @@
...
+            signature = value
+            continue
+        if key == 'ids[]':
+            if not ids:
+                query_params.append(('ids', ''))
+            ids.append(value)
+            continue
+        query_params.append((key, value))
+
+    message = urlencode(query_params, safe=':/').replace('ids=', f'ids={dumps(ids)}')
```
lishanl, line 45: I am trying to understand what is the effect of `safe=':/'` here. Some explanation would be helpful.

spylib/oauth/signature_validation.py
```diff
@@ -0,0 +1,81 @@
...
+    Returns:
+        A `SignedQueryString` instance.
+    """
+
+    signature: Optional[str] = None
+    query_params: List[Tuple[str, str]] = []
+
+    # I tried a number of options here to avoid doing
+    # manual parsing and string manipulation. `parse_qs`
+    # can work as long as you pass `doseq=True` & `quote_via=quote``
```
lishanl, line 28: is there an extra \` at the end of line? Instead of \`quote_via=quote`` -> \`quote_via=quote`?

lishanl:
> @hillairet @lishanl coming back to this, I am actually wondering if exporting anything other than `validate_*` is a mistake?
> 
> Maybe the parse fn could simply be merged into the validate one? That way we lower the API surface area? Let me know what you think!

Are you saying to just export `validate_signed_query_string` and do not export `SignedQueryString` and `parse_signed_query_string` via `spylib.oauth`? that sounds good, since `validate_signed_query_string` is probably mostly what the user needs. I wouldn't merge the parse fn into the validate one though. I prefer small and focused functions.

lishanl:
> Another point we need to clarify for the library as a whole. I'd say KISS, so we expose less first because it's simpler then we expose if needed. Removing/changing API causes pain to the user and docs annoyance for us so better to have less but more stability. We can expose more once we have reached some stability.

We could prefix the private methods / functions with `__` to indicate those are not to be accessed

qw-in:
> We could prefix the private methods / functions with `__` to indicate those are not to be accessed 

One of my bigger gripes with Python is that everything is "exported" by default. Wish there was like an `export` keyword or something

hillairet:
Single `_` would be enough. The double `__` is only enforced by the interpreter on class variables.

So moving the parsing to a classmethod, we'd have `_SignedQueryString` to indicate that it's internal and could change anytime. We could remove the underscore later on.

qw-in:
Ok, so I ended up just removing the parse fn - can always move the code back if it ends up being needed but the validate fn is a fine length imo.

Noticed in `oauth/validations.py` it attempts to handle sorting query string so I didn't want to remove that code and potentially break that behaviour. (seems like a workaround for a bonnette bug). Thoughts on how to handle?

lishanl (approved):
cool, thanks Quinn!

hillairet (approved):

PR #153, opened 2022-07-07
:sparkles: Document versioning 

### LINKED ISSUE #118: Add multiple version support to the documentation
It would be great to be able to go back to the documentation of older version, and not be restricted to the latest release.

ponty33:
close #118 
- Support versioning documentation
- Add step in github action workflow to push versioned document
- Use `fetch-depth: 0` as mentioned in [here](https://github.com/jimporter/mike/issues/49#issuecomment-809552399)
- Remove the step using `Deploy MkDocs` to deploy, as we are using mike, and keep that will remove the versioned docs
- This will auto trigger when new release is made, and create a new drop down according to the version tag
- I have accidently pushed the current 0.6.0 doc to the Github page, so you will be able to see the `0.6.0` dropdown on the top of the Spylib doc
- This screenshot is from a different repo I used to test Github Actions
Screenshot_20220721_111008

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -45,13 +45,13 @@ jobs:
         run: |
           source .venv/bin/activate
           lazydocs --overview-file="index.md" --src-base-url="https://github.com/SatelCreative/satel-spylib/tree/main" --output-path="./docs/api-docs" --no-validate spylib
-      - name: Release doc of new version
-      - run: |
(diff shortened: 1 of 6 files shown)


hillairet (approved):
Super sweet.
Why is it bad that you pushed 0.6.0?
I would say that it's good to have these older versions.
I wouldn't bother with 0.5.0 or prior but 0.6.0 is good to have!

lishanl (approved):
Thanks Frank.

PR #154, opened 2022-07-19
♻️ Fix/remove duplicated token models

### LINKED ISSUE #132: Merge the OnlineToken/OfflineToken response classes

lishanl:
fix #132 

- remove the duplicated models in the `admin_api` that's already extracted into `spylib/oauth/models.py` in PR #146 
- fix the associated tests

### DIFF
--- a/spylib/admin_api.py
+++ b/spylib/admin_api.py
@@ -31,30 +31,6 @@ from spylib.exceptions import (
 from spylib.utils.rest import Request
 
 
-class AssociatedUser(BaseModel):
-    id: int
-    first_name: str
-    last_name: str
-    email: str
-    email_verified: bool
-    account_owner: bool
-    locale: str
-    collaborator: bool
-
-
-class OfflineTokenResponse(BaseModel):
-    access_token: str
-    scope: str
-
-
-class OnlineTokenResponse(BaseModel):
-    access_token: str
-    scope: str
-    expires_in: int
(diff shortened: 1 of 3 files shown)


hillairet (commented):
Nice clean up!

lishanl:
> Nice clean up!

Is this supposed to be an approval instead of a comment? 😄 @hillairet

hillairet (approved):
Oops ... forgot to click "approve" :sweat_smile:

PR #155, opened 2022-07-22
:sparkles: Handle GQL non-200 response

### LINKED ISSUE #148: `execute_gql` does not check the response `status_code`
Title. It should probably make sure the response returns `200`

ponty33:
close #148 
[doc](https://shopify.dev/api/admin-graphql#status_and_error_codes)
- Raise exception if the response is not 200, Shopify GQL should always return 200
- According to the doc all non-200 errors will return in this format
```json
{
  "errors": "{Error message}"
}
```


lishanl (dismissed):
Thanks @ponty33 , it looks good. It would be great if you can add a test for it.

spylib/admin_api.py
```diff
@@ -187,6 +187,14 @@ async def execute_gql(
             headers=headers,
         )
         jsondata = resp.json()
+        # Handle any response that is not 200, which will return with error message
+        # https://shopify.dev/api/admin-graphql#status_and_error_codes
+        if resp.status_code != 200:
```
qw-in: Some responses may not contain valid JSON - we probably want to do the check before attempting to parse

hillairet:
@ponty33 Let's try to get this out of the way next week just to avoid stale PRs.

ponty33:
@hillairet Actually it's ready for review, I don't know why it's not passing the `tox`

hillairet:
Check the logs. That's for python 3.9:

image

hillairet (approved):

PR #156, opened 2022-07-25
:bug: Deprecate abstractclassmethod

### LINKED ISSUE #125: [mypy] Strange issue in mypy `0.950`
When updating to mypy `0.950`, a new error appears when writing the following code:

```python
from spylib import OfflineTokenABC

class OfflineTokenExecuter(OfflineTokenABC):
    async def save(self):
        pass

    @classmethod
    async def load(cls, store_name: str):  # <---  error is here
        # TODO: Loader function
        pass
```

`error: Signature of "load" incompatible with supertype "OfflineTokenABC"  [override]`

ponty33:
close #125 
- `abstractclassmethod` is deprecated, replace it with `classmethod` + `abstractmethod`

### DIFF
--- a/spylib/admin_api.py
+++ b/spylib/admin_api.py
@@ -1,5 +1,5 @@
 import logging
-from abc import ABC, abstractclassmethod, abstractmethod
+from abc import ABC, abstractmethod
 from asyncio import sleep
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


lishanl (approved):

hillairet (approved):
Thanks for clearing this!

PR #157, opened 2022-07-25
:sparkles: Extract the framework code

### LINKED ISSUE #135: Extract the framework agnostic code from init oauth router
There is a lot of code in `init_oauth_router` that has nothing to do with FastAPI and therefore should moved to another module or modules.

* [Lines 77 to 102](https://github.com/SatelCreative/spylib/blob/98154d88121347db475d4b6e922636e467ef4d3a/spylib/oauth/router.py#L77) could be something like `process_install_callback` (Maybe something better?)
* Similarly we should have a `process_login_callback` (Maybe something better?) with both `process_*_callback` functions using a function `process_callback` for [line 77 to 87](https://github.com/SatelCreative/spylib/blob/98154d88121347db475d4b6e922636e467ef4d3a/spylib/oauth/router.py#L77)

ponty33:
close #135 
- Move code to `callback.py` so nothing from Fastapi is included


lishanl (approved):

hillairet (approved):
Nice. But seeing that code again makes me want to make an abstract class!  :sweat_smile:

PR #163, opened 2022-08-13
💥🗑️ Remove bonnette specific code

### LINKED ISSUE #150: Merge new hmac signature validation with oauth validation if possible
While preparing a new validation function for hmac in PR #147, @qw-in noticed something ([original post](https://github.com/SatelCreative/spylib/pull/147#issuecomment-1167815252))

> Noticed in `oauth/validations.py` it attempts to handle sorting query string so I didn't want to remove that code and potentially break that behaviour. (seems like a workaround for a bonnette bug). Thoughts on how to handle?

@lishanl Could you make a branch, remove the sorting and test it out on azure somehow to find out if it's needed or not?
We don't need Bonnette anymore ...

lishanl:
Remove the sort on the OAuth args which was a fix for using [`bonnette`](https://github.com/jordaneremieff/bonnette) as the ASGI adapter for Azure Function. That is no longer needed and we can remove the corresponding code. 

Fix #150

I tested it with `st-cim` which no longer uses `bonnette` and it worked.

### DIFF
--- a/spylib/oauth/validations.py
+++ b/spylib/oauth/validations.py
@@ -1,5 +1,3 @@
-from copy import deepcopy
-from operator import itemgetter
 from typing import Any, List, Tuple
 from urllib.parse import parse_qsl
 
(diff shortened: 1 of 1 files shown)


hillairet (approved):
Nice clean up!

qw-in (approved):

PR #167, opened 2022-08-30
🐛 Fix fastapi tests

lishanl:
**Issue:** the CICD never ran the tests with `fastapi` installed because it's put in the dev-dependencies section in the `pyproject.toml` file and command `poetry install -E fastapi` wouldn't install it. 

**Fix:** move the `fastapi` to dependencies from dev-dependencies section and fix the broken tests


hillairet (approved):
Very nice!

PR #168, opened 2022-08-30
🎨 add bytes type data support to webhook hmac validate function

### LINKED ISSUE #164: Handle data to be provided as bytes in webhook validation
[This function](https://github.com/SatelCreative/spylib/blob/1b15af2abdbb3de8831e507f3097ee7f27e94a88/spylib/webhook/__init__.py#L26) just takes the data as a `str` which is a little annoying to have to convert from bytes in apps cause Starlette for example returns the body as `bytes`.
The dict/json version doesn't work because the ordering of the keys is not guaranteed.

Instead let's just have a check on the type and convert to a string using `.decode('utf8')` when the data is provided as bytes.

lishanl:
fix #164 

change webhook validate function to support both `str` or `bytes` data input 
from
```py
validate(data: str, hmac_header: str, api_secret_key: str)
```
to 
```py
validate(data: Union[str, bytes], hmac_header: str, api_secret_key: str)
```

### DIFF
--- a/spylib/webhook/__init__.py
+++ b/spylib/webhook/__init__.py
@@ -25,12 +25,15 @@ class WebhookCreate(Enum):
 
 
 def validate(data: Union[str, bytes], hmac_header: str, api_secret_key: str) -> bool:
+    data_str: str
     if isinstance(data, bytes):
-        data_str: str = data.decode(UTF8ENCODING)
+        data_str = data.decode(UTF8ENCODING)
     else:
-        data_str: str = data
+        data_str = data
     try:
-        validate_hmac(secret=api_secret_key, sent_hmac=hmac_header, message=data_str, use_base64=True)
+        validate_hmac(
+            secret=api_secret_key, sent_hmac=hmac_header, message=data_str, use_base64=True
+        )
     except ValueError:
         return False
     return True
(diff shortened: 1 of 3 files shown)


spylib/webhook/__init__.py
```diff
@@ -23,7 +24,9 @@ class WebhookCreate(Enum):
     PUB_SUB = 'pubSubWebhookSubscriptionCreate'
 
 
-def validate(data: str, hmac_header: str, api_secret_key: str) -> bool:
+def validate(data: Union[str, bytes], hmac_header: str, api_secret_key: str) -> bool:
+    if isinstance(data, bytes):
+        data = data.decode(UTF8ENCODING)
     try:
         validate_hmac(secret=api_secret_key, sent_hmac=hmac_header, message=data, use_base64=True)
```
hillairet: Let's be safe and let's not touch the argument
```suggestion
    if isinstance(data, bytes):
        data_str: str = data.decode(UTF8ENCODING)
    else:
        data_str: str = data
    try:
        validate_hmac(secret=api_secret_key, sent_hmac=hmac_header, message=data_str, use_base64=True)
```

hillairet (changes requested):
It won't change the argument outside of the function but I this it's best not to modify arguments, at least as a best practice for readability

hillairet (approved):
Coooool

PR #182, opened 2023-05-18
:sparkles: Change load to async

### LINKED ISSUE #169: Make load function in token abstract class async
For some reason this method was defined as sync instead of async [here](https://github.com/SatelCreative/spylib/blob/be7c3960e790166fd745bdfae4bccf1c6ca62707/spylib/admin_api.py#L280) and [here](https://github.com/SatelCreative/spylib/blob/be7c3960e790166fd745bdfae4bccf1c6ca62707/spylib/admin_api.py#L296)

And maybe try to see if mypy could run on the test folder so we can pick up this kind of silly mistake.

ponty33:
Close #169 
- Change `def` to `async def` for the load functions

### DIFF
--- a/docs/admin-api.md
+++ b/docs/admin-api.md
@@ -63,7 +63,7 @@ class OfflineToken(OfflineTokenABC):
         pass
 ```
 
-**Warning:** For version <= 0.8.1, async is **not** available for load function
+:warning: **Warning:** For version <= 0.8.1, async is **not** available for load function
 ```python
 class OfflineToken(OfflineTokenABC):
     @classmethod
--- a/docs/admin-api.md
+++ b/docs/admin-api.md
@@ -48,11 +48,6 @@ class OfflineToken(OfflineTokenABC):
     @classmethod
     async def load(cls, store_name: str):
         pass
-
-    # For version <= 0.8.1, async is not available for load function
-    @classmethod
-    def load(cls, store_name: str):
-        pass
 ```
 
 - If the Shopify API version used in Spylib is not the desired version, it can be changed with the following:
@@ -66,8 +61,11 @@ class OfflineToken(OfflineTokenABC):
     @classmethod
     async def load(cls, store_name: str):
         pass
+```
 
-    # For version <= 0.8.1, async is not available for load function
+**Warning:** For version <= 0.8.1, async is **not** available for load function
+```python
+class OfflineToken(OfflineTokenABC):
     @classmethod
     def load(cls, store_name: str):
         pass
--- a/docs/admin-api.md
+++ b/docs/admin-api.md
@@ -47,6 +47,7 @@ class OfflineToken(OfflineTokenABC):
 
   @classmethod
   async def load(cls, store_name: str):
+      # For version <= 0.8.x, use sync for load function
       pass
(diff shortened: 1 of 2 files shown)


hillairet (approved):

hillairet (changes requested):
Oh can you add a note in the docs actually? It's a breaking change!

ponty33:
The example in the doc is already written as `async`...
https://github.com/SatelCreative/spylib/blob/f191886c597b0a26f4911484b22a149865b26e49/docs/admin-api.md?plain=1#L19
Any suggestion on where else I should write the reminder?

hillairet:
> The example in the doc is already written as `async`...
> 
> https://github.com/SatelCreative/spylib/blob/f191886c597b0a26f4911484b22a149865b26e49/docs/admin-api.md?plain=1#L19
> 
> 
> Any suggestion on where else I should write the reminder?

At [this line](https://github.com/SatelCreative/spylib/blob/f191886c597b0a26f4911484b22a149865b26e49/docs/admin-api.md?plain=1#L36) you can something like:
*Note*: up to version 0.8.x, the load was sync instead of async.

Some sort of breaking change warning would be nice such as :boom:

lishanl (commented):
mmm maybe it's more like 
upto `0.8.1` it can be `def load` or `async def load`, see https://github.com/SatelCreative/kitchenhub/blob/main/deliverect-integration-app/deliverect-integration-app/shopify/client.py#L15-L20
after `0.8.1` it can only be `async def load` 
?

ponty33:
Prior to 0.8.1 should just be `sync`. I don't think you can override it with `async`

docs/admin-api.md
```diff
@@ -42,26 +42,36 @@ from spylib.admin_api import OfflineTokenABC
...
-  async def load(cls, store_name: str):
-      pass
-      
-# Add the version to be used here
-OfflineToken.api_version: '2022-01'
+    @classmethod
+    async def load(cls, store_name: str):
+        pass
+
+    # For version <= 0.8.1, async is not available for load function
```
hillairet: They should be separate to make it clearer that it's in different case and classes.
````suggestion
```

:warning: For version <= 0.8.1, async is not available for load function
```python
class OfflineToken(OfflineTokenABC):
````

**Review by hillairet — CHANGES_REQUESTED** [review]
Emphasize the difference

File: docs/admin-api.md [hunk 2]
```diff
@@ -42,26 +42,34 @@ from spylib.admin_api import OfflineTokenABC
...
-      pass
-      
-# Add the version to be used here
-OfflineToken.api_version: '2022-01'
+    @classmethod
+    async def load(cls, store_name: str):
+        pass
+```
+
+**Warning:** For version <= 0.8.1, async is **not** available for load function
```
hillairet: The emoji is even better to catch the eye
```suggestion
:warning: **Warning:** For version <= 0.8.1, async is **not** available for load function
```

hillairet (approved):

docs/admin-api.md
ponty33: Oh, didn't notice that's emoji, sure

PR #183, opened 2023-05-19
:sparkles: Remove default API version and make it manual input

### LINKED ISSUE #173: Update default API version.
[This](https://github.com/SatelCreative/spylib/blob/ed53a3175f5cbea84bec56e09a2b7416a017758b/spylib/admin_api.py#L45) is old. Maybe we should remove the default?

ponty33:
- Remove the `api_version` in `Token`, I couldn't found a way to override the field value
- Let `api_version` be input in `execute_rest` and `execute_gql`, since we construct full url inside the function
- If no version is entered, default to `/admin` which the oldest supported version will be used
- Modify test

close #173

### DIFF
--- a/tests/admin_api/test_initialization.py
+++ b/tests/admin_api/test_initialization.py
@@ -4,6 +4,7 @@ from ..token_classes import (
     OfflineToken,
     OnlineToken,
     PrivateToken,
+    VersionOfflineToken,
     offline_token_data,
     online_token_data,
     test_information,
@@ -58,4 +59,19 @@ async def test_private_token():
     assert (
         private_token.oauth_url
         == f'https://{test_information.store_name}.myshopify.com/admin/oauth/access_token'
+    )
+
+
+@pytest.mark.asyncio
+async def test_version_offline_token():
+    offline_token = await VersionOfflineToken.load(store_name=test_information.store_name)
+    print(offline_token.api_version)
+
+    assert offline_token.access_token == offline_token_data.access_token
+    assert offline_token.scope == offline_token_data.scope
+    assert offline_token.store_name == test_information.store_name
+    assert not offline_token.access_token_invalid
+    assert (
+        offline_token.api_url
+        == f'https://{test_information.store_name}.myshopify.com/admin/api/2023-04'
     )
--- a/tests/admin_api/test_initialization.py
+++ b/tests/admin_api/test_initialization.py
@@ -22,10 +22,7 @@ async def test_online_token():
     assert not online_token.access_token_invalid
     assert online_token.scope == online_token_data.scope
     assert online_token.associated_user_id == online_token_data.associated_user.id
-    assert online_token.api_url == (
-        f'https://{test_information.store_name}.myshopify.com/admin/'
-        + f'api/{test_information.api_version}'
-    )
+    assert online_token.api_url == f'https://{test_information.store_name}.myshopify.com/admin'
     assert (
         online_token.oauth_url
         == f'https://{test_information.store_name}.myshopify.com/admin/oauth/access_token'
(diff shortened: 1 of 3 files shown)


hillairet (commented):
I am very confused.aI thought I already commented on this! Did I forget to press send?

1. Do you really want to have to pass on the api version on every single call??? It would be annoying and error prone when it's option! Image some calls being on one API version and others on another version!
2. I don't understand why it is impossible to do `Token.api_version = '2023-04'`.  :thinking:

ponty33:
I will answer 2 first, yes, it just doesn't work. There are something about ABC behavior that I don't under stand, but I couldn't make it work when writing test for it
If 2 is not doable, the only work around I can think of is enter it in the arg
If you use the same constant such as `API_VERSION='2023-04'` in the code you can ensure it's always using the same version in the code. I know it's not pretty, but since simply assign doesn't work then I don't know what else I can do.

hillairet:
I don't see the problem here:
```python
from abc import ABC, abstractmethod
from typing import ClassVar, Optional

from pydantic import BaseModel


class Token(ABC, BaseModel):
    store_name: str

    api_version: ClassVar[Optional[str]] = None


class OfflineTokenABC(Token, ABC):
    @abstractmethod
    def version(self):
        pass


class MyOfflineTokenABC(OfflineTokenABC):
    api_version: ClassVar[Optional[str]] = '2023-04'

    def version(self):
        print(self.api_version)


offline = MyOfflineTokenABC(store_name='potato')
print(offline.api_version)
offline.version()

MyOfflineTokenABC.api_version = '2024-04'
print(offline.api_version)
print(Token.api_version)
```

Output as expected:
```
2023-04
2023-04
2024-04
None
```

So two ways of setting the api_version.
Maybe I'm not testing your problem though ...

ponty33:
I found where I did wrong...
When I wrote the test, I make the `load` function returning another class that doesn't have API version :sweat: 
No wonder the version is nowhere to be found :disappointed:

tests/admin_api/test_initialization.py
```diff
@@ -59,3 +60,18 @@ async def test_private_token():
         private_token.oauth_url
         == f'https://{test_information.store_name}.myshopify.com/admin/oauth/access_token'
     )
+
+
+@pytest.mark.asyncio
+async def test_version_offline_token():
+    offline_token = await VersionOfflineToken.load(store_name=test_information.store_name)
+    print(offline_token.api_version)
```
hillairet: You can test the value but print ain't pretty ...

tests/admin_api/test_initialization.py
```diff
@@ -59,3 +60,18 @@ async def test_private_token():
...
+
+    assert offline_token.access_token == offline_token_data.access_token
+    assert offline_token.scope == offline_token_data.scope
+    assert offline_token.store_name == test_information.store_name
```
hillairet: I'm not sure what you are testing other than the test code you wrote in `tests/token_classes` ...  :thinking:

tests/admin_api/test_initialization.py
```diff
@@ -59,3 +60,18 @@ async def test_private_token():
...
+    assert not offline_token.access_token_invalid
+    assert (
+        offline_token.api_url
+        == f'https://{test_information.store_name}.myshopify.com/admin/api/2023-04'
```
hillairet: Maybe the version could be a constant used in the class definition and this test?
```suggestion
        == f'https://{test_information.store_name}.myshopify.com/admin/api/{TEST_API_VERSION}'
```

hillairet (changes requested):
Small things

tests/admin_api/test_initialization.py
ponty33: Copy and paste issue :sweat_smile:

lishanl (approved):
🆒

hillairet (approved):
Neat

PR #190, opened 2023-06-09
✨ fastapi webhook hmac validate dependency

### LINKED ISSUE #165: Provide a fastapi dependency to validate the Shopify webhooks
We should be able to provide a dependency for fastapi/starlette users.

The challenge is that we don't just grab a header and check stuff on it. We need the whole request to get all the bits used to generate the webhook hmac.

One solution is to make our own [APIKeyHeader class](https://github.com/tiangolo/fastapi/blob/5c576e42be8f8832f0e225beaa72153a256892d7/fastapi/security/api_key.py#L41).

There is an example in our internal code [here](https://github.com/SatelCreative/st-pim/blob/7bf19af7a37c272a1339e5dda21d6ba781b05452/server/utils/authentication.py#L64).

It would probably look something like that:

```python
class HMACHeader(APIKeyBase):
    """Reuse APIKey to authenticate the webhook using the HMAC signature.
    It requires our own implementation of APIKeyHeader because we need to process the
    whole request object.
    """

    def __init__(self, *, name: str, scheme_name: str = None, auto_error: bool = True):
        self.model: APIKey = APIKey(**{'in': APIKeyIn.header}, name=name)
        self.scheme_name = scheme_name or self.__class__.__name__
        self.auto_error = auto_error

    async def __call__(self, request: Request) -> bool:
        return await validate_hmac(request)


key_hmac = APIKeyHMACHeader(name='X-Shopify-Hmac-SHA256')


def authenticate_hmac(hmac: bool = Security(key_hmac)) -> bool:
    if not hmac:
        raise HTTPException(status_code=HTTP_401_UNAUTHORIZED, detail='Webhook HMAC authentication failed')
    else:
        return hmac
```

lishanl:
close #165 

- add folder `fastapi_extensions`  instead of `fastapi` to avoid namespace conflict
- add `authentication.py` module for shopify webhook hmac validation to be used by fastapi dependency
- add tests
   - ignore the tests in `tests/fastapi_extensions` when fastapi is not installed


tests/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+def pytest_ignore_collect(path, config):
+    """Ignore the fastapi tests if fastapi is not installed.
+
+    pytest collection hook: https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_ignore_collect
+    """
+    if not fastapi_installed():
+        here = Path.cwd().absolute()
+        skip_fd = here / TO_IGNORE
+        if skip_fd == path:
+            return True
```
hillairet:
```suggestion
        return skip_fd == path
```

tests/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+def fastapi_installed() -> bool:
+    try:
+        from fastapi import FastAPI  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+def pytest_ignore_collect(path, config):
```
hillairet:
```suggestion
def pytest_ignore_collect(path, config) -> Optional[bool]:
```

spylib/fastapi_extensions/authentication.py
```diff
@@ -0,0 +1,77 @@
...
+
+try:
+    from fastapi import Request, Security  # type: ignore
+    from fastapi.openapi.models import APIKey, APIKeyIn  # type: ignore
+    from fastapi.security.api_key import APIKeyBase  # type: ignore
+except ImportError as e:
+    raise FastAPIImportError(
+        'The fastapi_extensions.authentication module require `fastapi` which is not installed. '
+        'Run `pip install spylib[fastapi]` to be able to use it.'
+    ) from e
```
hillairet: Aren't you doing the exact same thing in `__init__`?

spylib/fastapi_extensions/authentication.py
```diff
@@ -0,0 +1,77 @@
...
+def authenticate_webhook_hmac(hmac: bool = Security(webhook_hmac)):
+    """Dependency for authenticating the webhook using the HMAC signature.
+
+    `webhook_hmac.api_secret_key` needs to be assigned with a valid secret prior to using this dependency.
+
+    ```python
+    webhook_hmac.api_secret_key = 'SOME_SECRET'
+    @app.post('/SOME_PATH')
+        def authenticate(hmac: bool = Depends(authenticate_webhook_hmac)):
+            return hmac
```
hillairet:
```suggestion
    def some_path(is_hmac_valid: bool = Depends(authenticate_webhook_hmac)):
        if not is_hmac_valid:
            return HTTPException(status_code=400, detail='Invalid HMAC')
```

But now looking at the FastAPI docs, I wonder if we shouldn't drop support for 3.8 and use the [`Annotated` syntax](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-in-path-operation-decorators/#raise-exceptions) that seems much better.

hillairet (changes requested):
Two things that I realized just now:
1. We need also to handle public apps that are installed on multiple stores therefore have multiple API keys. That can be done by initializing the dependency with a loader function somehow. That's for another PR!
2. ~The dependency should send an exception and not return true/false cause otherwise the user always have to check and that reduces the benefits of the dependency injection greatly!~

Also maybe using 3.9 `Annotated` would be much better.

hillairet:
Ok I just saw that you are triggering an exception! My bad!
I guess we need to return something and the HMAC has no new info to return so bool is fine.

spylib/fastapi_extensions/authentication.py
lishanl: mm like using
1. [annotated metadata](https://fastapi.tiangolo.com/python-types/#type-hints-with-metadata-annotations)? Another PR?
2. The [annotated dependencies](https://fastapi.tiangolo.com/tutorial/dependencies/#share-annotated-dependencies) will require upgrade to fastapi [0.95.0](https://github.com/tiangolo/fastapi/releases/tag/0.95.0) as well. Another PR after the upgrade?

lishanl:
any comment @hillairet ?

hillairet (approved):

PR #192, opened 2023-06-21
:alien: 💥  replace app_handle with app_api_key for redirect

### LINKED ISSUE #181: Change final redirect in oauth
According to the [latest docs](https://shopify.dev/docs/apps/auth/oauth/getting-started#step-6-redirect-to-your-apps-ui) we should be redirecting to `https://{base64_decode(host)}/apps/{api_key}/
`.
That means we also don't need the app handle anymore though it still works.

lishanl:
close #181 

Do we want to also support redirect to non emebeded app like doing below?

```python
def app_redirect(
    store_domain: str, app_domain: str, app_api_key: str, embedded: bool = True
) -> str:
    if not embedded:
        return f'https://{app_domain}'
    return f'https://{store_domain}/admin/apps/{app_api_key}'
```

### DIFF
--- a/docs/oauth.md
+++ b/docs/oauth.md
@@ -31,7 +31,6 @@ oauth_router = init_oauth_router(
     private_key='KEY_FOR_OAUTH_JWT',
     api_key='SHOPIFY_APP_API_KEY',
     api_secret_key='SHOPIFY_APP_SECRET_KEY',
-    app_handle='SHOPIFY_APP_HANDLE',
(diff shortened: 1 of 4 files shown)


lishanl:
any comment / suggestion Ant @hillairet ?

hillairet (changes requested):
I do not see any change to the documentation ... :thinking:

hillairet (approved):
:rocket:

PR #197, opened 2023-07-22
⬆️ ➖ Update package dependencies

### LINKED ISSUE #176: Solve breaking changes on httpx 0.23.1 and fastapi 0.87.0
I have updated the `pyproject.toml` to not update the httpx and fastapi to the latest version because they both contain breaking changes. To be dealt with in another PR:
[httpx==0.23.1](https://github.com/encode/httpx/releases/tag/0.23.1), breaking change on url typing, likely to
[fastapi==0.87.0](https://github.com/tiangolo/fastapi/releases/tag/0.87.0), breaking change on testclient 

Please refer to the 2 CI failures in the PR.

_Originally posted by @lishanl in https://github.com/SatelCreative/spylib/issues/175#issuecomment-1350212603_

lishanl:
Update package dependencies to the latest (support the latest) possible and handled any changes required. 
Keep pydantic at v1 and fastapi under 0.100.* until pydantic v2 becomes stable.  
Remove `requests` from dev dependencies since the `fastapi` testclient uses `httpx`

close #176


pyproject.toml
```diff
@@ -13,38 +13,38 @@ repository = "https://github.com/SatelCreative/spylib"
...
-tenacity = "^7.0.0"
+httpx = ">= 0.18.1 <= 0.24.1"
+tenacity = "^8.2.2"
 starlette = ">= 0.15.0"
 pydantic ="^1.8.2"
 PyJWT = "^2.1.0"
 pycryptodome = "^3.10.1"
 
-fastapi = { version = ">= 0.69.0 < 0.87.0", optional = true }
+fastapi = { version = ">= 0.69.0 <= 0.99.1", optional = true }
```
hillairet: How about that instead?
```suggestion
fastapi = { version = ">= 0.69.0 < 0.100.0", optional = true }
```

hillairet (changes requested):
Could you ignore the specific error instead of all typing errors?
Only if there is one or two errors on a line. If it's a bunch then yeah use a blanket ignore.

hillairet (approved):

PR #200, opened 2023-08-08
:sparkles: Add basic embedded app

### LINKED ISSUE #180: Add basic embedded app just to have a fully installable app
Right now just using SPyLib leads to a 404 on the admin UI after finishing the install. Not great.
Ideally we'd have a dummy page good enough to render the app name or something basic like that.

ponty33:
close #180 
- Add a new arg `initial_path`. If set to `null` then there will be nothing on the root, so same 404 as what it is currently
- If any route is defined, it will then return a plain `App Installed`
- This endpoint would also check hmac so it can only be accessed by Shopify


ponty33:
@rahulpatidar0191 Do you know why the test is failing at `tox` step?

PR #205, opened 2023-12-01
:bug: Handle when the a Shopify API call response body is not a json

hillairet:
We have recently observed a weird response from Shopify that was not a JSON despite the GraphQL status code call being 200.
This ought to be an intermittent error on Shopify's side so we should retry the call to hopefully pass the problem.

I took the opportunity to parameterize the GraphQL error tests so I could add this case.
I also moved the number of retry attempts to a constant since it was in two locations.


lishanl (approved):
nice refactoring

PR #207, opened 2023-12-08
⬆️ Upgrade to pydantic 2.5

hillairet:
To get our projects to use pydantic 2, we need to also upgrade SPyLib so here we are!

Much easier than expected thanks to [bump-pydantic](https://docs.pydantic.dev/dev/migration/#code-transformation-tool) which made it easy.

When it couldn't do the change itself, it left a comment with instructions on what and how to do it. Very handy!

### DIFF
--- a/spylib/utils/jwtoken.py
+++ b/spylib/utils/jwtoken.py
@@ -50,6 +50,6 @@ class JWTBaseModel(BaseModel):
         -------
         The JWT as a string
         """
-        data = self.dict()
+        data = self.model_dump()
         data['exp'] = self.exp
         return jwt.encode(data, key, algorithm='HS256')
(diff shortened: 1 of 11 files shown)


spylib/utils/jwtoken.py
```diff
@@ -12,10 +14,11 @@ class JWTBaseModel(BaseModel):
     expiration time.
     """
 
-    exp: int = None  # type: ignore
+    exp: Optional[int] = None
 
-    @validator('exp', pre=True, always=True)
-    def set_id(cls, exp):
+    @field_validator('exp', mode="before")
```
lishanl:
```suggestion
    @field_validator('exp', mode='before')
```

lishanl (commented):
is that it? I expected more changes required. lol 😌

ponty33 (approved):

lishanl (approved):

PR #208, opened 2023-12-09
💥 Remove support for python version 3.8, add 3.12

hillairet:
In preparation of switching to pydantic version 2, we need to remove the support for python 3.8.
Pydantic 2 uses quite a lot `Annotated` in particular which was introduced in python 3.9 so it's just easier to drop 3.8.

I took this opportunity to add 3.12.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ pydocstyle = "^6.3.0"
 linkcheckmd= "^1.4.0"
 respx = "^0.20.2"
 mike = "^1.1.2"
+pytest-asyncio = "^0.23.2"
 
 [tool.poetry.extras]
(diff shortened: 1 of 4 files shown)


hillairet:
@lishanl PING

ponty33 (approved):
:wave: 3.8

pyproject.toml
```diff
@@ -30,7 +30,7 @@ pytest-mock = "^3.11.1"
 mypy = "^1.4.1"
 mypy-extensions = "^1.0.0"
 # latest blue 0.9.1 is not compatible with flake8 v5
```
lishanl:
```suggestion
```

lishanl (approved):

PR #210, opened 2023-12-09
⬆️ Upgrade packages in pyproject.toml

hillairet:
A lot of the dependencies were fixed to a specific version preventing proper upgrades. This is causing issues in the other PRs such as flake8 not working with python 3.12.

This upgrade will fix these issues and was needed anyway.
Blue doesn't seem maintained anymore since it doesn't support flake8 >= 5.0.0. So back to black!


ponty33 (approved):
:blue_heart: 
        v
:black_heart:

lishanl (approved):

PR #211, opened 2023-12-12
🐛  Shopify error handling

lishanl:
When Shopify GQL API returns 503 Service Unavailable, the content of the response is not JSON. 

The error being raise from AprilAire is 
```
  raise JSONDecodeError("Expecting value", s, err.value) from None
          │                                  └ '<!--\n  The source of truth for this generic error page is https://github.com/Shopify/error-pages.\n  Please do not edit thi...
          └ <class 'json.decoder.JSONDecodeError'>
```

Add retry to intermittent errors for 500 and 503 error codes
Add a try block for the json parsing to throw a more informative error for non 200 returns

### DIFF
--- a/spylib/admin_api.py
+++ b/spylib/admin_api.py
@@ -204,8 +204,8 @@ class Token(ABC, BaseModel):
 
             error_msg = jsondata['errors']
             raise ShopifyGQLError(
-                    f'GQL query failed, status code: {resp.status_code}. {error_msg}'
-                )
+                f'GQL query failed, status code: {resp.status_code}. {error_msg}'
+            )
 
         try:
             jsondata = resp.json()
--- a/spylib/admin_api.py
+++ b/spylib/admin_api.py
@@ -197,14 +197,15 @@ class Token(ABC, BaseModel):
         # Handle any response that is not 200, which will return with error message
         # https://shopify.dev/api/admin-graphql#status_and_error_codes
         if resp.status_code != 200:
-            jsondata = resp.json()
-            if jsondata:
-                error_msg = jsondata['errors']
-                raise ShopifyGQLError(
+            try:
+                jsondata = resp.json()
+            except JSONDecodeError:
+                raise ShopifyGQLError(f'GQL query failed, status code: {resp.status_code}.')
+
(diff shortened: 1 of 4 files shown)


spylib/admin_api.py
```diff
@@ -196,16 +199,22 @@ async def execute_gql(
...
-            else:
+            try:
+                jsondata = resp.json()
+            except JSONDecodeError:
                 raise ShopifyGQLError(f'GQL query failed, status code: {resp.status_code}.')
 
+            error_msg = jsondata['errors']
+            raise ShopifyGQLError(
+                f'GQL query failed, status code: {resp.status_code}. {error_msg}'
+            )
```
hillairet: issue: it feels like it can be simpler and more readable by highlighting the difference between having and not having the json data:
```suggestion
            try:
                jsondata = resp.json()
                error_msg = f'{resp.status_code}. {jsondata["errors"]}'
            except JSONDecodeError:
                error_msg = f'{resp.status_code}.'

            raise ShopifyGQLError(f'GQL query failed, status code: {error_msg}')
```

hillairet (changes requested):

hillairet (commented):
Wait, so you are handling this in the GraphQL but what about Rest?

lishanl:
> Wait, so you are handling this in the GraphQL but what about Rest?

rest has a broader retry covereage. 

https://github.com/SatelCreative/spylib/blob/73815ff5c06ecba5dc0e6bfb88085a4294bbd1b6/spylib/exceptions.py#L62-L67
https://github.com/SatelCreative/spylib/blob/73815ff5c06ecba5dc0e6bfb88085a4294bbd1b6/spylib/admin_api.py#L125-L131
https://github.com/SatelCreative/spylib/blob/73815ff5c06ecba5dc0e6bfb88085a4294bbd1b6/spylib/admin_api.py#L154-L156


https://github.com/SatelCreative/spylib/blob/73815ff5c06ecba5dc0e6bfb88085a4294bbd1b6/spylib/admin_api.py#L119-L121

hillairet (approved):
Fair enough!

PR #215, opened 2024-01-18
Fix/errors for tox

rahulpatidar0191:
Fix the [errors](https://github.com/SatelCreative/spylib/actions/runs/7574843332/job/20630548755#step:5:21) we were getting when running tests 
```  
  [RuntimeError]
  The Poetry configuration is invalid:
    - Additional properties are not allowed ('group' was unexpected)
  
  ```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - "**"
   pull_request:
     branches:
       - main
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - "**"
   pull_request:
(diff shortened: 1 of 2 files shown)


lishanl (approved):

hillairet (changes requested):
Nope.
The group is the new way for poetry to manage additional dependencies and the dev dependencies are deprecated so the fix ought to be on the tox side.
See the "DEPRECATED" in bold [here](https://python-poetry.org/docs/cli/#options-4).

lishanl:
> Nope. The group is the new way for poetry to manage additional dependencies and the dev dependencies are deprecated so the fix ought to be on the tox side. See the "DEPRECATED" in bold [here](https://python-poetry.org/docs/cli/#options-4).

🕵🏼

rahulpatidar0191:
Will fix in another PR

PR #216, opened 2024-01-24
🚀 Fix tox deployment errors

rahulpatidar0191:
- Fixes the [errors](https://github.com/SatelCreative/spylib/actions/runs/7574843332/job/20630548755#step:5:21) we were getting when running tests 
```  
  [RuntimeError]
  The Poetry configuration is invalid:
    - Additional properties are not allowed ('group' was unexpected)
  ```
- Removed duplicate installation of tox

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - "**"
   pull_request:
     branches:
       - main
@@ -27,7 +28,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install tox tox-gh-actions poetry
+          pip install tox-gh-actions poetry
 
       - name: Test with tox
         run: tox
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install tox-gh-actions poetry
+          pip install "tox-gh-actions>=3.2.0,<4.0.0" "poetry>=1.7.1,<1.8.0"
 
       - name: Test with tox
         run: tox
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - "**"
   pull_request:
     branches:
       - main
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by lishanl — CHANGES_REQUESTED** [review]
suggestion: let's also fix the versions for`tox-gh-actions` , `poetry`  and `tox` to whatever are the latest now so that we don't get any unexpected failure due to their breaking changes in the future latest versions.  Then we just regularly update their versions in a controlled manner.

**Comment by rahulpatidar0191** [comment]
> suggestion: let's also fix the versions for`tox-gh-actions` , `poetry` and `tox` to whatever are the latest now so that we don't get any unexpected failure due to their breaking changes in the future latest versions. Then we just regularly update their versions in a controlled manner.

Sure, updated `tox` and `poetry` `tox-gh-actions` is installed using github actions so I think it installs the latest version

File: pyproject.toml [hunk 1]
```diff
@@ -50,7 +50,7 @@ pytest-asyncio = "^0.23.2"
 fastapi = ["fastapi"]
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+requires = ["poetry-core>=1.7.1"]
```
lishanl:
```suggestion
requires = ["poetry-core>=1.8.1,<1.9.0"]
```

.github/workflows/tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install tox tox-gh-actions poetry
+          pip install tox-gh-actions poetry
```
lishanl:
```suggestion
          pip install "tox-gh-actions>=3.2.0,<4.0.0" "poetry>=1.7.1,<1.8.0"
```

lishanl: seems like poetry introduces breaking changes among minor update as well

pyproject.toml
lishanl: `poetry-core` sounds like a build too and different package from `poetry`, it too seem to introduce breaking changes among minor updates

lishanl:
Btw, this PR is to fix the poetry configuration and version mismatch when Tox runs the tests. not a deployment issue. 😄

lishanl (approved):
Thanks Rahul

rahulpatidar0191:
> Btw, this PR is to fix the poetry configuration and version mismatch when Tox runs the tests. not a deployment issue. 😄

It's the workflow that started this  😅

hillairet (approved):
Nice clean up.

PR #218, opened 2024-02-01
✨ Add connection test to token class

lishanl:


hillairet (commented):
thought: Should this return an object instead that could contain the response time later one?

lishanl:
> thought: Should this return an object instead that could contain the response time later one?

I guess it's about time not to ignore the speed of connection as well. 😅

docs/hmac.md
```diff
@@ -1,7 +1,7 @@
 # HMAC
 
 The hmac functionality can be used to validate the following Shopify requests:
- - [OAuth](https://shopify.dev/apps/auth/oauth/getting-started#step-7-verify-a-request)
+ - [OAuth](https://shopify.dev/docs/apps/auth/get-access-tokens/authorization-code-grant/getting-started#step-7-verify-a-request)
```
lishanl, line 4: Shopify actually redirects the old url to the new one but the linkcheck can't validate the redirect and returns 500

lishanl:
I want to reuse the time measurement for other connection test as well, so I made it a decorator in the `utils`

tests/oauth/test_base.py
```diff
@@ -89,15 +89,12 @@ async def test_oauth_with_fastapi(mocker):
...
+    # the url will change according to the pydantic version installed
+    del response_json['detail'][0]['url']
+    assert response_json == {
         'detail': [
-            {
-                'input': None,
-                'loc': ['query', 'shop'],
-                'msg': 'Field required',
-                'type': 'missing',
-                'url': 'https://errors.pydantic.dev/2.5/v/missing',
```
lishanl, line 99: the CI installs pydantic 2.6 and failed this test so just remove the url from the assert

lishanl:
any comment ant @hillairet ?

hillairet (approved):

hillairet (approved):
Nice change

hillairet:
@lishanl Here is an idea, just to get you thinking.
You could make a time class with the time stored in `_time` to show that it's not supposed to be accessed directly. Then add the properties `s()` and `ms()` this way you don't need to carry the unit in the name of the variable.

hillairet:
Or a less dumb idea: Use the `timedelta` class from `datetime`

lishanl:
> Or a less dumb idea: Use the `timedelta` class from `datetime`

I tried `timedelta` , it oddly doesn't have `milliseconds` attributes.

spylib/utils/misc.py
```diff
@@ -22,19 +22,31 @@ def parse_scope(v: Any) -> List[str]:
     return v
 
 
+class _Time(BaseModel):
+    seconds: float
+    milliseconds: float
```
hillairet: This class is not internal because TimedResult is not internal to this module. Plus I've never seen _ used on a class name. I wouldn't recommend it.
Also in English the "time" is a point in time so not clear name.
Finally, leverage the class functionalities to make your life easier and the usage more intuitive, at least to me.

```suggestion
class ElapsedTime(BaseModel):
    start: float
    end: float
    
    @property
    def seconds(self) -> float:
        return self.end - self.start
    
    @property
    def milliseconds(self) -> float:
        return 1000 * (self.end - self.start)
```

spylib/utils/misc.py
```diff
@@ -22,19 +22,31 @@ def parse_scope(v: Any) -> List[str]:
...
             result = await func(*args, **kwargs)
             end = perf_counter()
-            return data_type(result=result, elapsed_ms=(end - start) * 1000)
+
+            time_diff_seconds = end - start
+            return data_type(
+                result=result,
+                elapsed_time=_Time(
+                    seconds=time_diff_seconds, milliseconds=time_diff_seconds * 1000
+                ),
```
hillairet: Much better no?
```suggestion
                elapsed_time=ElapsedTime(start=start, end=end)
```

hillairet (changes requested):

spylib/utils/misc.py
lishanl: I thought what you meant by 
>You could make a time class with the time stored in _time

is make internal class. lol

hillairet:
> > Or a less dumb idea: Use the `timedelta` class from `datetime`
> 
> I tried `timedelta` , it oddly doesn't have `milliseconds` attributes.

Hummmmm I thought I saw it. :dizzy_face: 
One could extend the class to add this property but since the perf_counter is returning float time, timedelta doesn't make sense for this purpose anyway. So it was actually a dumber idea!  :sweat_smile:

hillairet (approved):

PR #221, opened 2024-02-21
:recycle: Refactor gql error handling

hillairet:
This part of the code was rather hard to follow and I think error prone although it was thoroughly tested.
This is however a nice cleanup that allows one to easily add more error handling without too much headache, I think. :pray: 
It's more lines of code but I believe that they are much more readable.


spylib/admin_api/gql_error_handler.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if not isinstance(jsondata, dict):
+            raise ValueError('JSON data is not a dictionary')
+
+        return jsondata
+
+    async def _check_errors_field(self, errors: list | str, jsondata: dict):
+        has_data_field = 'data' in jsondata
+        if has_data_field and not self.suppress_errors:
+            raise ShopifyGQLError(jsondata)
```
lishanl: thought/rhetoric question: would this scenario ever happen?

spylib/admin_api/gql_error_handler.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if isinstance(errors, str):
+            self._handle_invalid_access_token(errors)
+            raise ShopifyGQLError(f'Unknown errors string: {jsondata}')
+
+        await self._handle_errors_list(jsondata=jsondata, errors=errors)
+
+        errorlist = '\n'.join([err['message'] for err in jsondata['errors'] if 'message' in err])
+        raise ValueError(f'GraphQL query is incorrect:\n{errorlist}')
```
lishanl, line 90: question: check `suppress_errors` before raising just to keep the original logic? though I don't remember the purpose of `suppress_errors`. I have never set it to True.

spylib/admin_api/gql_error_handler.py
```diff
@@ -0,0 +1,154 @@
...
+
+    async def _handle_errors_list(self, jsondata: dict, errors: list) -> None:
+        # Only report on the first error just to simplify: We will raise an exception anyway.
+        err = errors[0]
+
+        if 'extensions' in err and 'code' in err['extensions']:
+            error_code = err['extensions']['code']
+            self._handle_max_cost_exceeded_error_code(error_code=error_code)
+            await self._handle_throttled_error_code(error_code=error_code, jsondata=jsondata)
```
lishanl, line 99: thought: it should be safe to assume the throttle error is the only error when it happens

lishanl (commented):
Thanks Ant. It's a lot more readable. 

Btw, python 3.9 doesn't support union with `|` that's what the python 3.9 codevalidation complained about. 
add the following to [enable future behavior](https://peps.python.org/pep-0563/#enabling-the-future-behavior-in-python-3-7 )? 
```py
from __future__ import annotations
```

spylib/admin_api/gql_error_handler.py
hillairet: I guess sometimes one can get data and errors ... not sure. The point here is not to change functionality so I didn't touch it.

spylib/admin_api/gql_error_handler.py
hillairet: Again, I reproduced the same functionalities and avoided making any changes.

spylib/admin_api/gql_error_handler.py
hillairet: That's a nice safety net, just in case. I don't know if it's possible but better to have code to catch it.

lishanl:
yea refactoring should preserve the original functionality / behavior.  👍🏼 Thanks ant

lishanl (approved):

PR #227, opened 2025-05-21
🔧 add setuptools to dev dependencies

lishanl:
The doc deployment is failing at the moment 
Poetry 2.0 doesn't come with `setuptools` which needs to be explicitly declared in the dev dependencies.
image


rahulpatidar0191 (approved):

lishanl:
Will need to assess a different library or github action to check markdown links.

PR #231, opened 2025-12-31
:alien: Oauth client credientials grant access token

lishanl:
Add to OfflineTokenABC class a new method `obtain_client_credentials_token` to get the access token by custom app client id and client secret


spylib/constants.py
```diff
@@ -11,3 +11,5 @@
 
 UTF8ENCODING = 'utf-8'
 API_CALL_NUMBER_RETRY_ATTEMPTS = 5
+
+TOKEN_EXPIRATION_BUFFER_SECONDS = 300
```
rahulpatidar0191, line 15: @lishanl What does this do?

rahulpatidar0191 (approved):

spylib/constants.py
lishanl: > @lishanl What does this do? 

Consider the token expired 5 min before the actual expiration (when used in the check) - a pretty generous time buffer for the calculation to detect token expiration for any renewal

